PR #4, opened 2015-04-01
postgresql storage provider

SakeSushiBig:
These commits add a new storage provider module for postgresql.
I tried to adopt as much from the mongodb implementation as possible, the only referenced I used.

Only requires one table in the specified database:
```sql
CREATE TABLE actor_sates (
   actor text NOT NULL,
   identity text NOT NULL,
   state text NOT NULL,
   PRIMARY KEY (actor, identity)
);
```
If the table does not exist the provider tries to perform a create statement.
The `state` column contains the specific actors state serialized in JSON.

Usage:
```java
OrbitStage stage = new OrbitStage();
final PostgreSQLStorageProvider storageProvider = new PostgreSQLStorageProvider();
storageProvider.setHost("localhost");
storageProvider.setPort(5433);
storageProvider.setDatabase("orbit");
storageProvider.setUsername("orbit");
storageProvider.setPassword("secret");
stage.addProvider(storageProvider);
// ...
```

Reading, writing, updating and clearing state are implemented & tested.

### DIFF
--- a/actors/providers/postgresql/pom.xml
+++ b/actors/providers/postgresql/pom.xml
@@ -29,65 +29,43 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <modelVersion>4.0.0</modelVersion>
     <parent>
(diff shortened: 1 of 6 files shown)


JoeHegarty:
Looks great. 
I think we'll need to disable the tests for each of the storage providers (Mongo, Postgres etc) and only run them if the developer wants to.

I'll look into merging this later today

Thanks for your contribution!

DanielSperry:
Hi Alexander. 
This is looking good! Thanks.
Using the json conversion was interesting. It avoids having to implement ORM.

JoeHegarty:
Merged. Thanks very much for your contribution.

You will need to run with the "withPostgresTests" using "-PwithPostgresTests" for the tests to run

PR #8, opened 2015-04-02
Linux shell scripts for the sample

voronaam:
Just a set of simple scripts to launch the sample on a Linux environment.


JoeHegarty:
Thanks a lot, that was on my to-do list.

PR #10, opened 2015-04-02
Scala sample PoC (don't accept, this is FYI)

voronaam:
This is a proof-of-concept that Orbit can work in Scala environment.

It is not good for anybody to really use it in production or in any other way. I only wanted to explore Scala compatibility and was delighted to see that all is needed is a custom macros.

I am running out of time I had to explore the framework - great job with it, btw! - meaning it is unlikely I will contribute scala macros today. If Scala is on your roadmap, you may look at this as a starting point.

You'd like Scala macros more that Java's AbstractProcessor. It operates on the AST instead of a StringBuffer with some source code...


DanielSperry:
Nice. I see that you're creating the reference factory by hand.
I'm just working on the runtime generation of those factories via bytecode.

It's in my personal fork: https://github.com/DanielSperry/orbit/commit/b647f4d6a5dddb9febab0de7dbfa59be571f448d

I'll be changing some names before merging it back to master. It will be there soon.
I think this will already work with scala.
Do you want to try it?

voronaam:
It should work via bytecode for any JVM language. I'll give your code a try, thank you!

DanielSperry:
I think I know what the problem is. Try adding this to your initialization:

     stage.setIncludedActors(Arrays.asList("ca.vorona.sample.trilean.*"));

It will instruct orbit to look at the classpath for actors.
This is not necessary with the code generation because that writes info to META-INF/orbit/...

DanielSperry:
I've got it working here. I'll merge and fit it into our project structure.

PR #15, opened 2015-04-07
Added Redis support

quilombodigital:
This is a preliminay redis support using jedis.


JoeHegarty:
Thanks for the contribution.
Your description only includes the redis provider but the pull request also contains "bridge".
Is that intentional?

quilombodigital:
Hi, 
Actually the redis provider is the first pull request. My intention was to send the bridge pull request later.
I´ve been using mercurial for a long time, and these are my first git adventures. I did not know that the following commits would be added to the current pull request.
Anyway, both commits are for your appreciation, but the bridge I consider a proof-of-concept. Just an idea of interoperability to be discussed by you, just test it locally and evolve the idea. :)

[]s, gandhi

JoeHegarty:
Thanks for your pull request. I've merged it and dropped the bridge from ml now, but have left the redis storage provider in there.
I like the idea of the bridge but introducing it will require some more discussion and thought.

quilombodigital:
Thanks a lot. I´ll try to learn a little more from git and pull requests.

And I agree totally that it should be discussed... heh, I can think about more two or three different ways to do it. :)

[]s, gandhi

JoeHegarty:
No problem, I'm also getting used to git, it's very different to the SCM I'm used to (Perforce).

PR #19, opened 2015-04-09
First implementation of InvokeInterceptor

quilombodigital:
This is a invoke interceptor that I created to allow external features to interfere with the method invocation without messing up with orbit core classes.

The main reason I implemented it is because I wanted to develop a custom Memoize annotation (result cache).
I have worked in some cluster configurations, and strategically placed memoized methods can save a *lot* of IO inside the cluster and increase the scalability. In a real world scenario there is always information that can be cached for a few seconds.

I added the memoize to the samples project because new annotations should be discussed, I´ve only added the interceptor support to the main project.

###### Usage

creating the stage:
```java
stage.setInvokeInterceptor(new MemoizeInterceptor());
```

actor:
```java
public interface IExample extends IActor {
    @Memoize(time = 5, unit = TimeUnit.SECONDS)
    Task<Long> getSomething(String greeting);
}
```

memoize interceptor example:
```java
public class MemoizeInterceptor implements InvokeInterceptor {

    ExpiringMap<String, Task> memoizeMap = ExpiringMap.builder().variableExpiration().build();

    @Override
    public Task invoke(IRuntime runtime, IAddressable toReference, Method m, boolean oneWay, int methodId, Object[] params) {
        if (m.isAnnotationPresent(Memoize.class))
        {
            Memoize ann = m.getAnnotation(Memoize.class);
            long memoizeMaxMillis = ann.unit().toMillis(ann.time());
            String key = Integer.toString(methodId) + "_" + Stream.of(params).map(p -> Integer.toString(p.hashCode())).collect(Collectors.joining("_"));
            Task cached = memoizeMap.get(key);
            if (cached == null)
            {
                cached = runtime.sendMessage(toReference, oneWay, methodId, params);
                memoizeMap.put(key, cached, ExpiringMap.ExpirationPolicy.CREATED, memoizeMaxMillis, TimeUnit.MILLISECONDS);
            }
            return cached;
        }
        return runtime.sendMessage(toReference, oneWay, methodId, params);
    }

}
```

In the future I would like to see a generic interceptor that could register multiple custom annotations with their respective lambdas....

[]s, Ricardo Mello (gandhi)


samples/hello/src/main/java/com/ea/orbit/samples/memoize/MemoizeInterceptor.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class MemoizeInterceptor implements InvokeInterceptor {
+
+    ExpiringMap<String, Task> memoizeMap = ExpiringMap.builder().variableExpiration().build();
+
+    @Override
+    public Task invoke(IRuntime runtime, IAddressable toReference, Method m, boolean oneWay, int methodId, Object[] params) {
+        if (m.isAnnotationPresent(Memoize.class))
```
johnou: minor, but instead of invoking isAnnotationPresent and then getAnnotation, you could simply call getAnnotation and then null check the result.

JoeHegarty:
Thanks. Looks good. I'll take a closer look at merging later today.

JoeHegarty:
Thanks for your submission. Invocation Hooks definitely seem valuable.

We've been reviewing this internally and while we agree that there is definitely a need for an invocation hook, we're wondering if this is the best way to implement it.

For similar cases in the framework we typically expose these kind of hooks via providers. I think this would easily translate into that model. 
The invocation hook would fit nicely into Execution with a new provider type (IInvocationHookProvider maybe).

You can take a look at ILifetimeProvider and it's use in Execution for an example of how this hooks into the framework.
e.g.
```java
Task.allOf(getAllProviders(ILifetimeProvider.class).stream().map(v -> v.preActivation(orbitActor))).join();
```

For the specific Memoize example, we're happy to take that but it should be a separate sample.
You could just copy the HelloWorld example and rename it and add the functionality there.

Please let me know if you have any questions, we'd like to get this into the framework.

quilombodigital:
Hi,

I modified it to work as you said, using an IInvokeHookProvider, Execution, and separated the memoize project.

At the same time, I understand that maybe you want to mature the idea. :)

The memoize project is not essential, can be ignored, but I think that it will eventually evolve to that CustomAnnotationHookProvider.

[]s, Ricardo Mello (gandhi)

quilombodigital:
Ah... just to make clear, dont worry about being polite in saying "no". :) Some pull requests like this are just excuses for feature proposals and to start healthy discussions. 8P

JoeHegarty:
Great. We'll take another look later today or tomorrow and try to get it merged.
Thanks!

DanielSperry:
Thanks for the contribution, @quilombodigital.

PR #26, opened 2015-04-14
Reminder support for SpringActorClassFinder

johnou:


actors/providers/spring/src/main/java/com/ea/orbit/actors/providers/spring/SpringActorClassFinder.java
```diff
@@ -75,7 +78,9 @@ public void initialize()
         {
             throw new NullPointerException("orbit.actors.basePackage must be set!");
         }
-        this.concreteImplementations = actorImplementationProvider.getActorImplementations(actorInterfaceProvider.getActorInterfaces());
+        Map<Class<?>, Class<?>> implementations = actorImplementationProvider.getActorImplementations(actorInterfaceProvider.getActorInterfaces());
+        implementations.put(IReminderController.class, ReminderController.class); // built in Orbit actor
```
johnou: maybe built-in Orbit system actors like ReminderController could be exposed somehow.

johnou: the latter makes more sense in my opinion.

DanielSperry: Indeed. The reminder controller initialization needs to be done from the nodes. 
Thanks.

DanielSperry:
Hi @johnou could you check if https://github.com/electronicarts/orbit/commit/c6ac3b3c8a9cf528b8a044cc268f34725855bdd7 addresses this issue?

PR #27, opened 2015-04-14
Cleanup

johnou:


actors/providers/redis/src/main/java/com/ea/orbit/actors/providers/redis/RedisStorageProvider.java
```diff
@@ -60,14 +56,13 @@ public RedisStorageProvider() {
...
 				.withSetterVisibility(JsonAutoDetect.Visibility.NONE)
-				.withSetterVisibility(JsonAutoDetect.Visibility.NONE)
 				.withCreatorVisibility(JsonAutoDetect.Visibility.NONE));
 		redis = new Jedis(host, port);
 		return Task.done();
 	}
 
 	private String asKey(final ActorReference reference) {
-		String classname = ActorReference.getInterfaceClass(reference).getName();
+		String classname = ActorReference.getInterfaceClass(reference).getSimpleName();
```
johnou: @quilombodigital let me know if this was intentional and I'll revert the revert :)

quilombodigital:
This is something I was in doubt and was planning to discuss at the chat. In my opinion, we should enforce all actors to have unique names across, and get rid of the package name. This would simplify all the keys we use at storage provider and other stuff.
This is a though thing to enforce.
So the answer is: the correct is to use getName(), to avoid name clashes, but I would like to use getSimpleName().

quilombodigital:
The question is that the actor identity is formed by the pair (class,instanceid). Another option would be to use another annotation to tell the "class" id, like @clusterid, but we cant solve everything using annotations. :)
We can talk more at the chat. :)

DanielSperry:
Thanks @johnou, we will be reviewing this soon.

PR #28, opened 2015-04-18
Added First Actor Method Trace implementation

quilombodigital:
Cluster Trace
======

tracedemo

This is a proof of concept for a Trace API for the orbit cluster
-----

I had the idea to develop a trace api to make the cluster flows easier to debug.
There were small changes to the core, most of the code here presented is of a demo application (*TraceDemoMain*) with views that use the trace api.
Some key decisions at the demo:
* Use a separated messaging system (simple multicast) to avoid using the cluster infra and affect the metrics.
* Possibility to disable tracing at the production
* Direct Listeners (not tasks) and queues for buffering.
* Support for receiver filtering and views.

There are a lot of possible improvements, but this should be discussed before more effort. Any doubts just drop a line. :)

[]s, Ricardo Mello (gandhi)

PS: I´ll be online rarely the next 5 days because of brazilian holydays.

### DIFF
--- a/actors/core/src/main/java/com/ea/orbit/actors/providers/IInvokeListenerProvider.java
+++ b/actors/core/src/main/java/com/ea/orbit/actors/providers/IInvokeListenerProvider.java
@@ -1,0 +1,16 @@
+package com.ea.orbit.actors.providers;
+
+public interface IInvokeListenerProvider extends IOrbitProvider
+{
+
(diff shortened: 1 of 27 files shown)


JoeHegarty:
This looks really nifty, thanks.
We should have time to go over this in more depth in the next couple of days.

DanielSperry:
@quilombodigital, this looks great! 
And you've managed to touch the orbit runtime very lightly to implement the trace.

JoeHegarty:
Merged. Thanks!

PR #30, opened 2015-04-19
Replace busy loop with simple latch.

johnou:


JoeHegarty:
Can't merge this right now, it appears there is some kind of deadlock. Perhaps @DanielSperry will have an idea why it doesn't work.

DanielSperry:
Hi @johnou this will not work when the application classes are loaded by a child ClassLoader.
This scenario happens for instance when we run `mvn test` in the project root.


After activation the orbit-async agent be loaded by the System ClassLoader.
Any orbit-async class previously loaded by the application class loader will be loaded again by the system class loader. 

In your case:

C1: Application-ClassLoader:AgentLoaderLatch
C2: System-ClassLoader:AgentLoaderLatch

C1 and C2 are different classes and share no data.

That's the reason for the deadlock. You are waiting on one latch and releasing the other.
I had the exact same problem when I was developing this.
I'll add a more detailed comment on the constant explaining this.

To share data between the two sets of `com.ea.orbit.async.instrumentation` classes,
we'd need to use a class that is certainly loaded by the system class loader.

That's why I've used the system properties. 
It's ugly but it's something guaranteed to happen only once.

We're open to suggestions on how the make this work differently.

samples/async/pom.xml
```diff
@@ -82,9 +82,6 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>2.18.1</version>
-                <configuration>
-                    <forkMode>never</forkMode>
```
DanielSperry: Removing this kinda hides the problem :p

JoeHegarty:
Going to close this one out given the comments by @DanielSperry.
Feel free to open another PR if you discover a nicer way to resolve it.

DanielSperry:
I found a way to use @johnou solution: first force the latch class to be loaded by the system class loader. And then invoke it's methods from the application using reflection.

Actually, it might be useful to do that to the entire AgentLoader.

PR #31, opened 2015-04-20
Fix timestamps in the frontend UI, make it work on OS X

pazaan:
- Fix timestamp rendering in the UI
- Add changes to start/stop scripts to support OS X
- Add log directory contents to .gitignore


JoeHegarty:
Looks Great. Thanks for your contribution!

PR #32, opened 2015-04-20
Adding logger to OrbitContainer and removing TODO

FieldFlux:
I'm not sure if INFO is appropriate here (maybe DEBUG?), or if we should be outputting any information here


JoeHegarty:
Looks good. Thanks!

PR #33, opened 2015-04-20
Fix observer manager cleanup.

johnou:
My assumption is that without ping in IActorObserver the method invocation is not done on the remote JVM?


JoeHegarty:
Ah, that one is my fault I think. I'll double check and merge if it's all good.

JoeHegarty:
Merged. Thanks!

PR #34, opened 2015-04-20
Fix observer manager cleanup.

johnou:


johnou:
On second thoughts I think this is a non-issue, please ignore this PR :)

PR #35, opened 2015-04-22
Fix compiler warnings.

johnou:
Should get rid of these warnings..

Warning:java: No SupportedSourceVersion annotation found on com.ea.orbit.async.processor.Processor, returning RELEASE_6.
Warning:java: Supported source version 'RELEASE_6' from annotation processor 'com.ea.orbit.async.processor.Processor' less than -source '1.8'


DanielSperry:
Thanks for the contribution, @johnou

PR #39, opened 2015-04-23
Added AnnotationHookProvider

quilombodigital:
Custom Annotation Provider
=======

This project substitute the memoize project for a more generic one, that can handle more annotations, since only one invokehook can be configured in Orbit.
------

The **@Memoize** and **@OnlyIfActivated** examples are available.

**Look at AnnotationTest.class for usage**


JoeHegarty:
I'll leave this one to you @DanielSperry as you understand the execution system better than I.

samples/annotation/src/main/java/com/ea/orbit/samples/annotation/onlyifactivated/OnlyIfActivatedAnnotationHandler.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public Class<OnlyIfActivated> annotationClass()
+    {
+        return OnlyIfActivated.class;
+    }
+
+    @Override
+    public Task<?> invoke(final OnlyIfActivated ann, final IRuntime runtime, final IAddressable toReference, final Method m, final boolean oneWay, final int methodId, final Object[] params)
+    {
+        INodeAddress address = runtime.locateActiveActor(toReference).join();
```
johnou: instead of using join here couldn't we use thenapply and null check address to avoid blocking?

quilombodigital:
Sure, this makes sense. The method is local now... We dont know tomorrow :). I´m at my job now, later I can do this.

quilombodigital:
This hook is not exactly what I wanted yet. 
It is a little complicated how annotations can affect each other when we are talking about execution flow. For example, Memoize and OnlyIfActivated annotations cant be used together currently, and I could not see an easy way to do it without mapping their interaction.
Which one should be executed first? Respect Annotation order? Return on first match?
My first idea was to chain (compose) the tasks, but I dont know if it is a good idea, so I kept the minimum to work with multiple annotations.

DanielSperry:
This is looking promising. I've just merged it.
Thanks for the contribution @quilombodigital !

DanielSperry:
I think this show that we need to chain the hooks and let each modify the input and intercept the result of the next. 

@johnou, this will make for instance the Memoize work with the OnlyIfActive.

I'll do something on that direction on the Execution later.

PR #42, opened 2015-04-24
Memcached storage provider.

johnou:


actors/core/src/main/java/com/ea/orbit/actors/runtime/OrbitActor.java
```diff
@@ -120,9 +121,9 @@ protected T state()
      *
      * @return a completion promise
      */
-    protected Task<Boolean> readState()
+    protected Task<Void> readState()
```
DanielSperry: Could you document the information conveyed by the Boolean in the return type?

DanielSperry: Cool, you removed the unused boolean. I wonder if it shouldn't signal if the data was found or not.

quilombodigital:
Yes, I found it interesting too, but the question is, what the application can do with the boolean besides logging or trying again?

quilombodigital:
Real world scenario, the database goes down... the actors keep running.... the database goes up again, hum... I guess that only actors that were deactivated with a failed writestate will loss data. Can we avoid somehow postpone actors deactivation when there were writestate failures?

DanielSperry:
Yes, but that turns one problem into another: memory.

quilombodigital:
Ok, so make it a choice. The system by default keeps the actor activated until a valid writestate. And the user can add a special provider that can choose in what situations the actors can be deactivated despite the writestate success. Another approach would be to configure a max memory limit for this situation, where it will deactivate the oldest in this situation when a memory limit is reached.

actors/providers/memcached/README.md
```diff
@@ -0,0 +1,4 @@
+Orbit Actors Storage Provider Memcached
+============
+
+Here be dragons, the Memcached persistence test invokes "flush_all". Running it against a shared Memcache instance will result in data loss.
```
DanielSperry: scary

DanielSperry:
There was a reason not to accept the provider's data object into the actor.
In the future we'd like to be able to manage the actor state.
That's something that for instance Orleans does.

In another life and company I implemented a transactional system that also monitored all access to the application objects. So this would be a simple thing to do. I'm not sure if it's a good thing, though. Now I come to think of this sort of monitoring as an early optimization bound to be more expensive than just cloning the thing.

So while we decide if the object will or will not be updated in place.
@johnou have you looked into libraries like [Dozer](http://dozer.sourceforge.net/) or [kryo](https://github.com/EsotericSoftware/kryo) they make it possible to clone or copy objects.

What would allow the in place replacement of object would be something like:

```
Task<Boolean> readState(ActorReference<?> reference, Object state) {
    Object o = readFromSomeDatabase(...);
    mapper.copy(o, state);
    return done();
}
```

I understand your concern with asynchronous reads (where the application neglects to `await(readState());` )
However if the application chooses to use async reads, it will be in trouble with the atomic state as well.
Because it will still be possible that two subsequent calls to `state()` will return different objects.
I agree that that is better than the object fields changing mysteriously during the same call.


So as you can see this discussion is not over yet. 
And I'd really appreciate to hear your arguments.

johnou:
@DanielSperry thanks for such a detailed insight of what you might have planned for the future. To be honest I hadn't thought as far as monitoring application objects and I was not happy with changing the API of the storage provider so I will go ahead and revert those changes. 

Dozer looks reasonable in terms of updating the actor state object so I will update the Memcached provider to use that.

I also updated the test to just loop all known keys, inefficient sure, but not dangerous.

actors/providers/memcached/src/main/java/com/ea/orbit/actors/providers/memcached/MemCachedStorageProvider.java
```diff
@@ -0,0 +1,117 @@
...
+ * @author Johno Crawford (johno@sulake.com)
+ */
+public class MemCachedStorageProvider implements IStorageProvider
+{
+
+    private Mapper mapper;
+    private MemCachedClient memCachedClient;
+    private MemCachedStorageHelper memCachedStorageHelper;
+
+    private boolean useShortKeys = false;
```
johnou: Useful for innodb backed memcached storage.

actors/providers/memcached/src/main/java/com/ea/orbit/actors/providers/memcached/ByteUtils.java
```diff
@@ -0,0 +1,129 @@
...
+            {
+                enableResolveObject(true);
+            }
+            @Override
+            protected Object resolveObject(final Object obj) throws IOException
+            {
+                if (obj instanceof ReferenceReplacement)
+                {
+                    ReferenceReplacement replacement = (ReferenceReplacement) obj;
+                    // TODO: help?
```
johnou: @DanielSperry any suggestions on what we can do from this context?

actors/providers/postgresql/src/main/java/com/ea/orbit/actors/providers/postgresql/PostgreSQLStorageProvider.java
```diff
@@ -55,7 +55,7 @@
     private String username;
     private String password;
 
-    private Connection conn;
+    private Connection connection;
     private PreparedStatement insertState;
```
johnou: general observation of the postgres implementation, I would consider removing the sync keywords and creating a new prepared statement every time we need to read / write and close it in a finally block..

actors/providers/memcached/src/main/java/com/ea/orbit/actors/providers/memcached/ByteUtils.java
DanielSperry: `IActor.getReference` if the class is an IActor.
else `IActorObserver.getReference()` which doesn't exist yet. I'll add it shortly.

DanielSperry:
@johnou I've added two utility methods for restoring references:  https://github.com/electronicarts/orbit/commit/18c036a62347e7434d74b676d4e489af99f49676

JoeHegarty:
Merged, thanks Johno.

DanielSperry:
:+1:

PR #45, opened 2015-05-01
Improved Invocation Hook to use a chain of Providers

quilombodigital:
@DanielSperry and I did remote pair programming tonight and this is the result. :)
hook
This is an improved invocation hook that allows a better control of the invocation flow, making possible customs annotations to work together, and invocation features to be implemented as extensions, keeping the orbit core clean.


DanielSperry:
That photo tells me that it's a good thing we don't have a human child together :)

PR #46, opened 2015-05-05
Added support for mutiple storages by using StorageProvider annotation

quilombodigital:
This is a minimum support for multiple storages. The developer just have to set a storage name and use the same name using the annotation.
-------

```
RedisStorageProvider storageProvider = new RedisStorageProvider();
storageProvider.setName("gamestate")
storageProvider.setTimeout(60000);
storageProvider.setDatabaseName(databaseName);
stage.addProvider(storageProvider );

...

@StorageProvider(name = "gamestate")
public class SomeActor extends OrbitActor<SomeActor.State> implements ISomeActor
{
}
```

Note: The default behaviour is mantained, without any annotation the first provider is used.
-------


DanielSperry:
This is looking good @quilombodigital!
Thanks for the contribution!

PR #48, opened 2015-05-09
Ldap Storage Provider

quilombodigital:
LDAP Storage Provider
=======

###### How to use it

Adding the provider:

```java
        LdapStorageProvider provider = new LdapStorageProvider();
        provider.setHost("localhost");
        provider.setPort(10389);
        provider.setDn("uid=admin,ou=system");
        provider.setCredentials("secret");
        stage.addProvider(provider);
```

State example:

```java
@LdapEntity(
        dnKey = "uid",
        baseDn = "ou=people, dc=example, dc=com",
        attributes = {
                "ObjectClass: inetOrgPerson",
                "ObjectClass: organizationalPerson",
                "ObjectClass: top",
                "ObjectClass: person"
        })
public class LdapUserState
{

    @LdapAttribute("cn")
    public String commonName;

    @LdapAttribute("sn")
    public String surName;

    public String postalCode;

    public String userPassword;

}
```

Actor Example:

```java
public class LdapAuthenticateActor extends OrbitActor<LdapUserState> implements ILdapAuthenticateActor
{

    @Override
    public Task<Boolean> authenticate(final String password)
    {
        if (!readState().join() || state().userPassword == null)
        {
            return Task.fromValue(false);
        }
        return Task.fromValue(state().userPassword.equals(LdapStorageUtil.toPasswordHash(password)));

    }

    @Override
    public Task<Void> register(final String name, final String surname, String postalcode, final String password)
    {
        state().postalCode = postalcode;
        state().commonName = name;
        state().surName = surname;
        state().userPassword = LdapStorageUtil.toPasswordHash(password);
        writeState().join();
        return Task.done();
    }

    @Override
    public Task<Void> remove()
    {
        clearState().join();
        state().userPassword = null;
        state().commonName = null;
        state().postalCode = null;
        state().surName = null;
        return Task.done();
    }

}
```

**Note1: The user authentication example was chosen because it is the most common scenario, but different approachs can be used.**
**Note2: I used locally the ApacheDS(apacheds-2.0.0-M14) and Apache Directory Studio(2.0.0-M8) to test the provider.**

### DIFF
--- a/actors/providers/jpa/src/main/java/com/ea/orbit/actors/providers/jpa/JpaStorageProvider.java
+++ b/actors/providers/jpa/src/main/java/com/ea/orbit/actors/providers/jpa/JpaStorageProvider.java
@@ -158,8 +158,4 @@ public class JpaStorageProvider extends AbstractStorageProvider
         }
     }
 
-    private String getIdentity(final ActorReference<?> reference)
-    {
-        return String.valueOf(ActorReference.getId(reference));
-    }
 }
--- a/actors/providers/ldap/pom.xml
+++ b/actors/providers/ldap/pom.xml
@@ -66,10 +66,10 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
             <version>${project.version}</version>
         </dependency>
         <dependency>
-			<groupId>org.apache.directory.shared</groupId>
-			<artifactId>apache-ldap-api</artifactId>
-			<version>1.0.0-M13</version>
-		</dependency>
+            <groupId>org.apache.directory.api</groupId>
+            <artifactId>api-all</artifactId>
+            <version>1.0.0-M30</version>
+        </dependency>
         <dependency>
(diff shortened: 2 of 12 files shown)


JoeHegarty:
Seems there is a problem with the dependencies and the build fails.
Otherwise it looks good.

Can you resolve the below issue?

Failed to execute goal on project orbit-actors-ldap: Could not resolve dependencies for project com.ea.orbit:orbit-actors-ldap:jar:0.2.4-SNAPSHOT: Failure to find org.apache.directory.shared:apache-ldap-api:jar:1.0.0-M13

Thanks!

quilombodigital:
I´ve fixed the maven dependency. Thanks. :)

JoeHegarty:
Perfect, thanks. Merged

DanielSperry:
This is very creative! Wow :)

DanielSperry:
@quilombodigital Do you think it's possible to automate the testing of this with travis?

quilombodigital:
I´ll check later and try to make a test, but probably needs a openldap install

PR #49, opened 2015-05-10
Added embedded apacheds LDAP server for tests

quilombodigital:
I initially tried to make a travis shell script to install openldap, but it was a total mess to configure. I eventually discovered that apacheds already has an integration with unit tests and I use it.
I´m not totally in favor of using the server that shares code with the client api, I think that by using a different server, it would be more easy to find compatibility bugs.
I also enabled the test at the travis file, I dont know if it will be enabled by default.

### DIFF
--- a/actors/providers/ldap/pom.xml
+++ b/actors/providers/ldap/pom.xml
@@ -65,11 +65,13 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
             <artifactId>orbit-actors-json</artifactId>
             <version>${project.version}</version>
         </dependency>
+
         <dependency>
-            <groupId>org.apache.directory.api</groupId>
-            <artifactId>api-all</artifactId>
-            <version>1.0.0-M30</version>
+            <groupId>org.apache.directory.server</groupId>
+            <artifactId>apacheds-all</artifactId>
+            <version>2.0.0-M20</version>
         </dependency>
+
         <dependency>
             <groupId>com.ea.orbit</groupId>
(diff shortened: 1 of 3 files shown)


JoeHegarty:
Merged. Thanks.

actors/providers/ldap/pom.xml
```diff
@@ -65,11 +65,13 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
             <artifactId>orbit-actors-json</artifactId>
             <version>${project.version}</version>
         </dependency>
+
         <dependency>
-            <groupId>org.apache.directory.api</groupId>
-            <artifactId>api-all</artifactId>
-            <version>1.0.0-M30</version>
+            <groupId>org.apache.directory.server</groupId>
+            <artifactId>apacheds-all</artifactId>
```
johnou: instead of including apacheds-all with the entire ldap module it could be limited to test scope, right? then perhaps api-ldap-client-api for the ldap module itself to keep it lean.

quilombodigital: Actually this is what I tried first @johnou :), but there is an issue with the apacheds and api-ldap-client-api where apacheds finds "multiple schemas" at the classpath and fails to run. I tried to use maven exclusion to solve the issue, but did not worked. After several tries I just gave up, but you can give a try. :)

quilombodigital: The error was something like this: http://stackoverflow.com/questions/17217465/apache-directory-ldap-api-getting-up-and-running

PR #52, opened 2015-05-19
Jpa Storage fix to support objects not marked as entities

quilombodigital:
This is a fix for a recent issue, caused because the Jpa Provider only supports entity objects.
Now when an unsupported object is used, it is converted into a json object and persisted using a generic jpa object. 
A test was also created to check the reminder persistence at the providers.


JoeHegarty:
Thanks Ricardo. Looks good!

PR #59, opened 2015-05-28
Actor Stage JMX Initial Implementation

FieldFlux:
First pass at exposing some Actor Stage metrics via JMX.

With this PR, stages will expose Local Actor count, Observer Instance Count, Number of messages received, handled, and the number of refused executions. 

It publishes these metrics via JMX. You can use jconsole to see these values and you can also use a variety of tools to move this data into other tools like Graphite.


JoeHegarty:
Thanks, I'll look at merging this tomorrow.

DanielSperry:
Modifying the stage (by adding the JMX interface) is a bit drastic. 
We could move this code to an extension.

PR #61, opened 2015-05-31
Ensure change state notifications are sent before shutting down the executor.

johnou:
Ensure change state notifications are sent before shutting down the executor.

Motivation:

When shutting down an Orbit node the STOPPED state is never sent.

Modifications:

During shutdown sequence we now wait for the change state notifications to be sent before shutting down the executor.

2015-05-31 12:49:41,213 ERROR [ForkJoinPool-2-worker-441] (com.ea.orbit.actors.runtime.Messaging) Error processing message.
java.util.concurrent.RejectedExecutionException
	at java.util.concurrent.ForkJoinPool.fullExternalPush(ForkJoinPool.java:1531)
	at java.util.concurrent.ForkJoinPool.externalPush(ForkJoinPool.java:1501)

Result:

Improved graceful shutdown of Orbit nodes.

### DIFF
--- a/actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
+++ b/actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
@@ -478,7 +478,7 @@ public class Execution implements Runtime
     {
         // * refuse new actor activations
         state = NodeCapabilities.NodeState.STOPPING;
-        hosting.notifyStateChange();
+        hosting.notifyStateChange().join();
 
         // * deactivate all actors
         activationCleanup().join();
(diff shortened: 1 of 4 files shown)


johnou:
@DanielSperry I assume the reason why com.ea.orbit.actors.runtime.Messaging#stop is commented out in com.ea.orbit.actors.runtime.Execution#stop and com.ea.orbit.actors.runtime.ExecutionSerializer#shutdown doesn't wait for shutdown is because of unit tests? Looks as though some pending requests may be discarded..

DanielSperry:
Yes that was one of the reasons. Did you get it to work now?

johnou:
@DanielSperry yes, assuming the next build passes ;) https://github.com/electronicarts/orbit/pull/65

PR #65, opened 2015-06-01
Shutdown and wait for messaging.

johnou:


actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
```diff
@@ -498,7 +498,7 @@ public void setMessaging(final Messaging messaging)
         Task.allOf(extensions.stream().map(v -> v.stop())).join();
 
         // * cancel all pending messages, and prevents sending new ones
-        //messaging.stop();
+        messaging.stop();
```
johnou: @JoeHegarty do you know why @DanielSperry left this commented?

JoeHegarty:
I'm not 100% sure why that code doesn't run. Daniel should be around tomorrow so I'll let him answer before merging.

actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
johnou: @DanielSperry wouldn't it make sense to stop the storage provider and other extensions after stopping messaging?

actors/test/actor-tests/src/main/java/com/ea/orbit/actors/test/ActorBaseTest.java
```diff
@@ -198,7 +206,8 @@ protected boolean isIdle(Stage stage)
         {
             // this is very ad hoc, but should work for our tests, until execution changes.
             // for starters access to this map should be synchronized.
-            Map running = (Map) getField(getField(getField(stage, "execution"), "executionSerializer"), "running");
+            Map running = (Map) getField(getField(getField(stage, Stage.class, "execution"), Execution.class,
```
johnou: bit more obvious like this and if we ever use an anonymous inner class to create ExecutionSerializer reflection will continue to work.

actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
DanielSperry: I think that was commented because it was creating problems with some tests. 
Is it working now?

PR #66, opened 2015-06-01
Using the intended cluster name instead of the default one

Fiurgeist:
l. 172: adding call to TransportConfigurationBuilder.clusterName
- using the clusterName defined in the orbit.yaml instead of the default Infinispan one
- as discussed in Gitter, using the default one was not done on purpose, so I suggest setting the intended one

l. 151: removed UUID call
- since ForkChannel.connect ignores the parameter one provides, appending a random UUID to it is superfluous and a waste of processing power


DanielSperry:
@Fiurgeist thanks for the contribution.

PR #67, opened 2015-06-02
Minor jpa refactoring fix

quilombodigital:
minor fix to to last refactoring


JoeHegarty:
Ah, missed that one. Thanks

PR #68, opened 2015-06-02
Stage lifecycle

johnou:
@DanielSperry something like this? You may want to repackage the lifecycle class, had a hard time trying to find it a good home.


DanielSperry:
Look good :)

DanielSperry:
But why make it mandatory?

johnou:
I wouldn't call an interface with a default method mandatory? There are not going to be many extensions, the call is not expensive, only invoked on shutdown and it reduces boilerplate code in the extension implementation itself.

DanielSperry:
good point :)

PR #69, opened 2015-06-06
Orbit Metrics

FieldFlux:
A metrics layer based on dropwizard's metrics library. It is still pretty simple and only exposes Gauge functionality and not any of the more interesting features of Dropwizard Metrics like histograms. 

Currently, objects need to be manually registered with the MetricsManager (see Stage.java for an example). Eventually I'd like to see this change in some fashion. Any fields or methods tagged with @ExportMetrics will be registered for with the Metrics system. 

Current Implementation supports SL4J, Graphite, and Ganglia reporting and multiple reporters can be active at once.

Configuration takes place in orbit.yaml like so:

```
orbit.metrics.reporters:
  - !!com.ea.orbit.actors.metrics.config.reporters.Slf4jReporterConfig
    period: 1
    periodUnit: 'MINUTES'
  - !!com.ea.orbit.actors.metrics.config.reporters.GraphiteReporterConfig
    period: 1
    periodUnit: 'MINUTES'
    host: 127.0.0.1
    port: 2003
```

Feedback is appreciated! :)


JoeHegarty:
This looks really great. I'll take a closer look and get some feedback from others before proceeding with merging

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
```diff
@@ -77,6 +79,9 @@
     @Wired
     Container container;
 
+    @Inject
```
JoeHegarty: We should avoid doing this as if you aren't running in a container or DI framework it won't work.

JoeHegarty:
In general it may make sense for us to put this in core, and not rely on any injection etc. That way both container and actors can use it, as well as any other components in future.

FieldFlux:
Yeah I had thought about putting it in commons, but I was concerned about adding the dependencies to commons since it currently doesn't have any. I can easily make that move if you'd prefer :)

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
FieldFlux: Hmm, If we aren't running the container the @Config line would break as well wouldn't it since we rely on the Container to inject the config?

JoeHegarty: Yes, that's correct. In other cases where that might happen, we have the @Config and we provide manual getter/setter for those not running in container. You can see that with the Configs in stage, and how we set up the samples which aren't running container (Hello).

With @Inject it will just be null not running in container, so we should probably provide a static getting or similar and use that. I assume dropwizard is thread safe, so we'll just need to be sure our wrapper is.

JoeHegarty:
I think in this case it makes sense to be in commons. We don't like to introduce dependencies in there, but I think this is a good exception to the rule.

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
johnou: Instead of the metrics manager approach wouldn't it be more flexible and modular to use the extensions system? Would also be easier to setup for those not using Orbit container.

actors/metrics/src/main/java/com/ea/orbit/actors/metrics/MetricsManager.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        for (Field field : obj.getClass().getDeclaredFields())
+        {
+            if (field.isAnnotationPresent(ExportMetric.class))
+            {
+                //check to see if the field is accessible.
+                if (!field.isAccessible())
+                {
+                    continue;
```
johnou: Log warning?

actors/metrics/src/main/java/com/ea/orbit/actors/metrics/MetricsManager.java
```diff
@@ -0,0 +1,148 @@
...
+                }
+                final ExportMetric annotation = field.getAnnotation(ExportMetric.class);
+                final String gaugeName = MetricRegistry.name(obj.getClass(), annotation.name());
+
+                registry.register(gaugeName, (Gauge<Object>) () -> {
+                    try
+                    {
+                        Object value = field.get(obj);
```
johnou: Could extract this into its own method registerGauge and reuse it below?

actors/metrics/src/main/java/com/ea/orbit/actors/metrics/MetricsManager.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        for (Method method : obj.getClass().getDeclaredMethods())
+        {
+            if (method.isAnnotationPresent(ExportMetric.class))
+            {
+                //methods declared as metrics must not accept parameters.
+                if (method.getParameterCount() > 0)
+                {
+                    continue;
```
johnou: Log warning, programming error + silent fail = bad.

actors/metrics/src/main/java/com/ea/orbit/actors/metrics/MetricsManager.java
```diff
@@ -0,0 +1,148 @@
...
+                    {
+                        throw new IllegalStateException("Method " + method.getName() + " was inaccessible: " + iae.getMessage());
+                    }
+                    catch (InvocationTargetException ite)
+                    {
+                        throw new UncheckedException("Invocation of method " + method.getName() + " failed", ite.getTargetException());
+                    }
+                });
+
+                logger.debug("Registered new metric for method " + method.getName() + " in class " + obj.getClass());
```
johnou: Not really any performance benefit here but I generally recommend to wrap debug calls with isdebugenabled.

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
```diff
@@ -241,6 +246,15 @@ public void setMode(final StageMode mode)
         hosting.start();
         execution.start();
 
+        if (metricsManager != null)
+        {
+            String cleanRuntimeId = runtimeIdentity().replaceAll("[\\[\\]\\.\\\\/]", ""); //strip illegal characters
+            cleanRuntimeId = cleanRuntimeId.replace("Orbit","");
+            cleanRuntimeId = getClusterName() + "." + getNodeName() + "." + cleanRuntimeId;
```
johnou: I'd suggest moving the cleaning into the metrics manager init method.

actors/metrics/src/main/java/com/ea/orbit/actors/metrics/MetricsManager.java
```diff
@@ -0,0 +1,148 @@
...
+{
+    static final MetricRegistry registry = new MetricRegistry();
+    private static final Logger logger = LoggerFactory.getLogger(MetricsManager.class);
+    private boolean isInitialized = false;
+
+    @Config("orbit.metrics.reporters")
+    private List<ReporterConfig> reporterConfigs;
+
+
+    public synchronized void initializeMetrics(String uniqueId)
```
johnou: I would still prefer the extensions approach but here I would throw illegal state exception or at least log warning if the manager is initialised multiple times.

johnou:
As a bonus it would be nice to incorporate the long counter LongAdderV8 like Netty have here https://github.com/netty/netty/commit/6eac574b0e9b0be5212c56c5c4e21aad001160dd

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
JoeHegarty: I think the metrics manager approach actually makes more sense in this context. It means it's usable from any Orbit module. 
Extension also doesn't make a lot of sense as this metrics need to be integrated into the core to be effective.

FieldFlux:
If we go ahead with moving metrics into commons, does it still make sense for Stage to be in charge of initializing the reporters? That pretty much means that a Stage will always be required to be started before metrics can be piped out. What if there are multiple stages within a Container? Metrics would report under the Id of whichever stage happened to start first. 

Perhaps Metrics should be a Startable and initialized seperately?

JoeHegarty:
My preference would be to have MetricsManager be a regular Java object that is safe to call init on multiple times.

Then have:
* A startable wrapper that calls MetricsManager.init in commons
* Call MetricsManager.init in Stage as well

This is similar to how we do Async, so it gets initialized if you are only using container or actors (and still works if you are using both).

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
johnou: Then we should probably move the metrics package out of actors to the base?

johnou:
I like the idea of integrating metrics like async so it is an optional dependency, metrics could be a compile time dependency on stage allowing for annotations, and we could use class.forname for init. This would allow us to keep metrics in its own module like async.

@DanielSperry instead of registering metrics manually what do you think about extending the class path scanner? or creating a new annotation scanner to allow for automatic registration. 

@JoeHegarty any ideas for configuration in a non orbit container environment, extract to its own config file or set before manager init?

JoeHegarty:
Yes, maybe making metrics be an optional package is the best choice, and using the the class.forname type stuff like we did with Async would make the most sense.
It could fit nicely in it's own module in utils/metrics.

For config I'd suggest that we allow setting before the first init somehow. We'd need to be careful about init order though.

FieldFlux:
I like the idea of mirroring Async. I'll work on getting that up and running in the next day or so. :)

FieldFlux:
We could put the annotations into commons though :)

FieldFlux:
So the StatelessActor Test exposes a potential problem with the Singleton MetricsManager approach. It attempts to startup multiple stages, the first of which intiializes metrics with it's runtimeId. Subsequent  stage starts attempt to register metrics but because metrics an only be seeded once, metric registration beings to fail due to duplicate uniqueIds.

Perhaps MetricsManager should not be a singleton and each Stage should start it's own instance?

JoeHegarty:
Does it need a unique id?
I'm just wondering if the rule should be "one metrics manager per JVM". I'm thinking about a case where you're running both actors and container (and potentially something else) which all use metrics manager.

FieldFlux:
each metric has to have a unique id. It's an implementation decision by dropwizard. With this particular case, since more than one stage is being started in a JVM, multiple instances of the com.ea.orbit.stage.execution.localactorcount metric is added to the single JVM wide MetricsManager. Currently I'm using class name + member name to build my unique IDs which obviously isn't unique enough

I'm planning on adding support for appending a unique runtime id to metrics so that we can support actor integration, but if it is intended to be a 1:1 relationship between stage and JVM then perhaps we should disable metrics on those specific tests.

Alternatively we can introduce the concept of a metric set. That way individual components or stages can group up their metrics to reduce potential collisions.

Thoughts?

JoeHegarty:
Hmm, I think we need to be able to support multiple instances of the same metric with some kind of runtime id. As if we wanted to instrument say AbstractActor wouldn't we have the same problem?
Perhaps the metrics set idea is also the best option for stages as they really are a logical grouping.

What is your preference to solve it?

quilombodigital:
I did not saw youir implementation, but this is my opinion for this discussion:
I´m currently working on some metrics for my servers, and I dont think the system should add any stage runtime id to the metric name. That is what "Reporters" do, by using their own configuration.
This is what collectd graphite plugin does when sending to influxdb and grafana, the machine metrics are the same.
For example, the dropwizard graphiteReporter supports '.prefixedWith("orbit.node.A")'
https://dropwizard.github.io/metrics/3.1.0/manual/graphite/#manual-graphite
About the actors support, I see no problem by using an actor name to deal with actor wide metrics, and an id/number as part of the metric name for an actor instance. That would work ok in grafana.

johnou:
Depends on your use case for Actor metrics but for us it has been easier to inject a statistics singleton. It gets quite messy with jmx registration depending on mbean policies and when actors deactivate and possibly move to another node.

FieldFlux:
Currently, I'm using the reporters prefix method to prefix metrics with <ClusterName>.<NodeName>.<CleanedStageRuntimeId>.blah just to keep things organized a bit. I'll need to rework this naming convention now that metrics can be started and used outside of the Actors module. The easiest solution would be to pull it out to the config to make it easier, but I was hoping to do something at runtime to ease automated deployment of nodes. Currently the Chat example fires up two "backend-stage" nodes which doesn't really help in differentiating between them. Hence the StageId being added. I know this is more of an Ops issue, but are there any plans to attack this?

I think that given that the common use is one stage per JVM, we can simply accept the fact that metrics initialization will fail. I need to catch the exception, log it and fail gracefully anyway. 

Thinking about it more, I think that in situations where we need to collect metrics that are common to all instances of a particular Actor, the singleton approach is probably best. Injection works, but maybe another Actor could be created whose sole job is to act as a sink for common metrics to a particular actor might work (this might work around the issue of actors moving between nodes). For stats that can be keyed to the specific actor instance, we can append the actor's unique id to the metric. Using this approach requires minor changes to my implementation and with sufficient design work ahead of time for those using the metrics API, provides a workable solution.

If, in the future we decide we need metrics sets, it can be added to the API, but I think right now, just reworking things a bit we can get away without it.

Using Graphite as an example since I'm most familiar with it, I'm picturing something like this hierarchy:

<configurablePrefix>.com.ea.orbit.actors.stage.execution.localActorCount

<configurablePrefix>.ca.fieldflux.awesomeapp.FooActor.commonmetric1 

<configurablePrefix>.ca.fieldflux.awesomeapp.FooActor.<ActorId>.uniqueMetric

I think this is a more appropriate structure. The configurable prefix can be changed via the orbit.yaml file and possibly can be altered by automated deploy systems. Does this sound agreeable?

johnou:
We addressed this by setting a service identifier per JVM via system property defaults to service-type-dev.

```code
    {
      "@class": "org.jmxtrans.embedded.output.GraphiteWriter",
      "settings": {
        "namePrefix": "xxxx.${service.env}.${service.name}.",
        "host": "${jmxtrans.graphite.host}",
        "port": "${jmxtrans.graphite.port}",
        "enabled": "${jmxtrans.graphite.enabled}"
      }
    }
``

JoeHegarty:
Hey @FieldFlux, do you want me to leave this PR open?
No rush if you're busy, just curious where we're at with this.

FieldFlux:
Yeah, been fighting some fires here. I need to make a few changes to gracefully handle duplicate metric registrations and resolve the new conflict and then I think we're probably pretty good for a first-pass implementation. We can close and I can open a new PR if you'd like.

JoeHegarty:
Ok. I'll do that so we have a fresh PR to review once you are done.
Thanks for all the effort, we really appreciate it.

PR #73, opened 2015-06-20
Orbit Metrics Round 2

FieldFlux:
This PR provides a wrapper around Dropwizard's metrics library.

Fields or Methods tagged with @ExportMetric can be registered to emit metrics to a variety of sources via Reporters. In this PR, wrappers are provided for emitting metrics to SL4J, JMX, Graphite, and Ganglia. Reporters can be configured via orbit.yaml like so:

```
orbit.metrics.reporters:
  - !!com.ea.orbit.metrics.config.Slf4jReporterConfig
    period: 1
    periodUnit: 'MINUTES'
    prefix: 'chat.chat-backend'
  - !!com.ea.orbit.metrics.config.JmxReporterConfig {}
```
Standard configuration management tools can be used to modify the prefix on a host-by-host manner when deploying by modifying this file. 

An Orbit Container Module is provided to setup metrics on the JVM if actors is not being used. The Actors Module will also attempt to setup metrics. If Container is not being used, MetricsManager can be provided a set of ReporterConfig objects using the initializeMetrics() method. This should be done before Stage startup if actors is being used. 

Metrics must be registered using the MetricsManager registerExportedMetrics method. A instanceId may be provided so that instance metrics can be setup. 

Metrics must be uniquely named. Currently, metrics are named using the name of the class they are defined in plus the name provided via the ExportMetric annotation. For example a metric in ca.fieldflux.awesomeapp.neatclass with a field annotated with 
```
@ExportMetric(name="coolmetric")
```
would generate a metric named: 
```
ca.fieldflux.awesomeapp.neatclass.coolmetric
```

Instance Metrics can be used where there may be more than one instance of a class and metrics must be tracked per-instance. To mark a metric as an instance metric use:
```
@ExportMetric(name="coolmetric", isInstanceMetric=true)
```
and provide an instanceId when registering the object instance. This will insert an instance identifier into the metric name hierarchy like so:
```
ca.fieldflux.awesomeapp.neatclass.instanceId.coolmetric
```

An Actors lifetime extension is also provided so that Actors can expose metrics and have them automatically registered and unregistered on actor activation and deactivation respectively.


johnou:
Instead of isInstanceMetric how about "singleton" and default to true?

utils/metrics/src/main/java/com/ea/orbit/metrics/MetricsManager.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    public static MetricsManager getInstance()
+    {
+        return instance;
+    }
+
+    public static String sanitizeMetricName(String name)
+    {
+        return name.replaceAll("[\\[\\]\\.\\\\/]", ""); //strip illegal characters
```
johnou: Extract into static pattern / compile regex once?

actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsLifetimeExtension.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class MetricsLifetimeExtension implements LifetimeExtension
+{
+    @Override
+    public Task preActivation(AbstractActor actor)
+    {
+        try
+        {
+            Class.forName("com.ea.orbit.metrics.MetricsManager"); //make sure the metrics manager is on the classpath.
+            String actorId = ActorReference.getId(ActorReference.from(actor)).toString();
```
johnou: Unfortunately this isn't going to work, we cannot have imports for MetricsManager or reference it directly if you want the dependency to be optional, instead we'll need to use reflection to instantiate the manager.

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
```diff
@@ -241,6 +246,18 @@ public void setMode(final StageMode mode)
         hosting.start();
         execution.start();
 
+        try
+        {
+            Class.forName("com.ea.orbit.metrics.MetricsManager"); //make sure the metrics manager is on the classpath.
+
+            MetricsManager.getInstance().initializeMetrics(metricsConfig);
```
johnou: Unfortunately this isn't going to work, we cannot have imports for MetricsManager or reference it directly if you want the dependency to be optional, instead we'll need to use reflection to instantiate the manager.

FieldFlux: hmm, well thats certainly going to make things uglier...

JoeHegarty: Yeh, it won't be super pretty but you should be able to achieve it with reflection.
That's the only major problem I see with the code, otherwise it looks great!

FieldFlux: Should be fixed in latest push :)

FieldFlux:
I apparently reformatted Stage.java by mistake. Will attempt to fix.

FieldFlux:
Should be better now. :)

JoeHegarty:
Looks good. I'll merge this shortly if there are no further comments :)

actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsLifetimeExtension.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ */
+public class MetricsLifetimeExtension implements LifetimeExtension
+{
+    @Override
+    public Task preActivation(AbstractActor actor)
+    {
+        try
+        {
+            Class.forName("com.ea.orbit.metrics.MetricsManager"); //make sure the metrics manager is on the classpath.
```
johnou: imo redundant check, if the user imports this class they should have the metrics module in the classpath, should be safe to remove the forname try catch.

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
```diff
@@ -404,4 +414,40 @@ public void bind()
...
+            Class mmClazz = Class.forName("com.ea.orbit.metrics.MetricsManager"); //make sure the metrics manager is on the classpath.
+            Method getInstanceMethod = mmClazz.getDeclaredMethod("getInstance");
+            Method initializeMetricsMethod = mmClazz.getDeclaredMethod("initializeMetrics", List.class);
+            Method registerExportedMetricsMethod = mmClazz.getDeclaredMethod("registerExportedMetrics", Object.class, String.class);
+
+            Object managerObject = getInstanceMethod.invoke(null, null);
+            initializeMetricsMethod.invoke(managerObject, metricsConfig);
+            registerExportedMetricsMethod.invoke(managerObject, execution, execution.runtimeIdentity());
+        }
+        catch (Exception ex)
```
johnou: instead of catching exception how about two catches, one for class not found == metrics isn't being used and another to catch other errors == log error.

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
```diff
@@ -404,4 +414,40 @@ public void bind()
...
+            Class mmClazz = Class.forName("com.ea.orbit.metrics.MetricsManager"); //make sure the metrics manager is on the classpath.
+            Method getInstanceMethod = mmClazz.getDeclaredMethod("getInstance");
+            Method unregisterExportedMetricsMethod = mmClazz.getDeclaredMethod("unregisterExportedMetrics", Object.class, String.class);
+
+            Object managerObject = getInstanceMethod.invoke(null, null);
+            unregisterExportedMetricsMethod.invoke(managerObject, execution, execution.runtimeIdentity());
+        }
+        catch (Exception ex)
+        {
+            //OK. Just means that metrics isn't being used.
```
johnou: see above.

utils/metrics/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+
+       <dependency>
+           <groupId>io.dropwizard.metrics</groupId>
+           <artifactId>metrics-graphite</artifactId>
+           <version>${metrics.version}</version>
+       </dependency>
+
+       <dependency>
+           <groupId>io.dropwizard.metrics</groupId>
+           <artifactId>metrics-healthchecks</artifactId>
```
johnou: healthchecks in use?

utils/metrics/src/main/java/com/ea/orbit/metrics/MetricsComponent.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Config("orbit.metrics.reporters")
+    private List<ReporterConfig> metricsConfig = new ArrayList<>();
+
+    @Override
+    public Task<?> start()
+    {
+        try
+        {
+            Class.forName("com.ea.orbit.metrics.MetricsManager"); //make sure the class is in the classpath.
```
johnou: redundant, should be in the classpath.

utils/metrics/src/main/java/com/ea/orbit/metrics/MetricsManager.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.regex.Pattern;
+
+public class MetricsManager
+{
+    private static MetricsManager instance = new MetricsManager();
+    private static final MetricRegistry registry = new MetricRegistry();
+    private static final Logger logger = LoggerFactory.getLogger(MetricsManager.class);
+    private Map<ReporterConfig, Reporter> reporters = new HashMap<>();
+    private boolean isInitialized = false;
+    private static Pattern nameSanitizationRegex = Pattern.compile("[\\[\\]\\.\\\\/]");
```
johnou: Not 100% sure of EAs coding conventions but maybe final and uppercase for constant?

utils/metrics/src/main/java/com/ea/orbit/metrics/annotations/ExportMetric.java
```diff
@@ -0,0 +1,43 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD, ElementType.METHOD})
+public @interface ExportMetric
+{
+    String name() default "";
+
+    boolean isInstanceMetric() default false;
```
johnou: still keen for the term singleton but it's your call.

FieldFlux: As discussed on Gitter, What if we used an enum like so:
```
@ExportMetric(name="metricname", scope=Singleton)
@ExportMetric(name="instanceMetric", scope=Instance)
```

Except that @johnou  is suggesting that we use "Prototype" instead of "Instance." Do you have any thoughts @JoeHegarty?

FieldFlux:
I have removed the redundant checks for Class.ForName in a local commit. Will push after we come to a decision on the naming of instance metrics :)

JoeHegarty:
The naming you guys discussed on gitter looks good to me.

@ExportMetric(name="metricname", scope=Singleton)
@ExportMetric(name="metricname", scope=Prototype)
@ExportMetric(name="instanceMetric", scope=Instance)

JoeHegarty:
@FieldFlux if you want to push those last remaining items, I'll merge this today.

utils/metrics/src/main/java/com/ea/orbit/metrics/annotations/ExportMetric.java
JoeHegarty: The naming you guys discussed on gitter looks good to me.

@ExportMetric(name="metricname", scope=Singleton)
@ExportMetric(name="metricname", scope=Prototype)
@ExportMetric(name="instanceMetric", scope=Instance)

FieldFlux:
Instance metrics are now "Prototype Metrics" and the scope is controlled via an enum on the ExportMetrics annotation.

PR #75, opened 2015-06-24
Add documentation of pending response behaviour + unit test.

johnou:


actors/stage/src/main/java/com/ea/orbit/actors/runtime/Messaging.java
```diff
@@ -69,8 +70,8 @@
     private ClusterPeer clusterPeer;
     private Execution execution;
     private final AtomicInteger messageIdGen = new AtomicInteger();
-    private Map<Integer, PendingResponse> pendingResponseMap = new ConcurrentHashMap<>();
-    private PriorityBlockingQueue<PendingResponse> pendingResponsesQueue = new PriorityBlockingQueue<>();
+    private final Map<Integer, PendingResponse> pendingResponseMap = new ConcurrentHashMap<>();
+    private final PriorityBlockingQueue<PendingResponse> pendingResponsesQueue = new PriorityBlockingQueue<>(50, new PendingResponseComparator());
```
johnou: 50 seems like a reasonable number for initial capacity (required to use the comparator constructor), default in JDK source is 11..

actors/stage/src/main/java/com/ea/orbit/actors/runtime/Messaging.java
```diff
@@ -93,27 +94,39 @@ public NodeAddress getNodeAddress()
...
             {
-                return messageId - o.messageId;
+                return o1.messageId - o2.messageId;
             }
             return cmp;
         }
+    }
 
-        @Override
-        public boolean equals(final Object o)
```
johnou: Since the equals / hashCode contract is broken just to keep FindBugs happy I extracted the compare to a Comparator.

PR #82, opened 2015-07-06
Misc cleanup.

johnou:


actors/stage/src/main/java/com/ea/orbit/actors/runtime/cloner/KryoCloner.java
```diff
@@ -56,18 +57,17 @@ public KryoCloner()
         // exception.
```
johnou: I'd be interested in knowing what exception is thrown.

JoeHegarty:
Thanks @johnou

PR #84, opened 2015-07-09
Add setter for messaging to allow configuring response timeout via st…

johnou:
…age.


johnou:
Not sure how I missed this, using reflection to configure it in the meanwhile :)

PR #85, opened 2015-07-21
Correctly detect unmodifiable collection.

johnou:
I think it is clearer like this and now unmodifiableCollection is detected (which I believe is what Chris originally wanted).


JoeHegarty:
Thanks @johnou

PR #87, opened 2015-07-22
Cache refactor.

johnou:


actors/stage/src/main/java/com/ea/orbit/actors/runtime/cloner/KryoCloner.java
```diff
@@ -40,75 +43,76 @@
...
-            {
-                try
-                {
-                    return kryo.copy(obj);
-                } finally
-                {
-                    kryo.reset();
-                }
-            }
+            return kryoPool.run(kryo -> kryo.copy(object));
```
johnou: Unsure if we should be invoking reset here explicitly, at least run takes care of returning the kyro instance back to the pool.

PR #91, opened 2015-07-28
Orbit Web Metrics Integration

FieldFlux:
I've reorganized the folder structure for Metrics to move it down a level. I've then added another project  to allow for integration of Dropwizard's Jetty handler instrumentation.

I opted to put the Jetty stuff into orbit-metrics-web so that Orbit Metrics doesn't need a dependency on jetty.


JoeHegarty:
Thanks Jeff!

PR #95, opened 2015-08-16
DynamoDB Storage Extension

JoeHegarty:
This is an actor state extension for Amazon's DynamoDB storage solution.
This is the first fully asynchronous storage extension so as such it's still very experimental.

The PR includes the extension itself, tests and Travis-CI integration.


DanielSperry:
Thanks @JoeHegarty this is looking good!

PR #96, opened 2015-08-21
Stage shouldn't initialize Metrics

FieldFlux:
Its redundant and shouldn't be Stage's responsibility. Users not using Container can initialize Metrics themselves. Users using Container have a handy Component that does it for them.

### DIFF
--- a/utils/metrics/metrics/src/main/java/com/ea/orbit/metrics/MetricsComponent.java
+++ b/utils/metrics/metrics/src/main/java/com/ea/orbit/metrics/MetricsComponent.java
@@ -36,10 +36,13 @@ import com.ea.orbit.metrics.config.ReporterConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.inject.Singleton;
+
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 
+@Singleton
 public class MetricsComponent implements Startable
 {
     private static final Logger logger = LoggerFactory.getLogger(MetricsComponent.class);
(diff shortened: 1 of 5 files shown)


utils/metrics/metrics/src/main/java/com/ea/orbit/metrics/MetricsManager.java
```diff
@@ -100,9 +103,9 @@ public synchronized void initializeMetrics(List<ReporterConfig> reporterConfigs)
         }
         else
         {
-            if (logger.isDebugEnabled())
+            if (logger.isWarnEnabled())
             {
-                logger.debug("Attempting to initialize the Metrics Manager when it is already initialized!");
+                logger.warn("Attempting to initialize the Metrics Manager when it is already initialized!");
```
FieldFlux: @JoeHegarty I've put this back to a WARN as it now is a suspicious state that should be flagged. This change prevents the spam seen in tests anyway :)

utils/metrics/metrics/src/main/java/com/ea/orbit/metrics/MetricsComponent.java
```diff
@@ -36,10 +36,13 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.inject.Singleton;
+
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 
+@Singleton
```
FieldFlux: Without this, apparently Container wouldn't start the Component. Is that intended?

utils/metrics/metrics/src/main/java/com/ea/orbit/metrics/MetricsManager.java
JoeHegarty: Sounds good

utils/metrics/metrics/src/main/java/com/ea/orbit/metrics/MetricsComponent.java
JoeHegarty: Yes, that's normal. Only actual singleton components will be started by the framework on it's own, otherwise it doesn't know what the lifetime should be.

FieldFlux: ah right. Makes sense :)

PR #101, opened 2015-10-09
Migrating to Infinispan 8.0.1.Final

DanielSperry:


actors/stage/src/main/java/com/ea/orbit/actors/cluster/JGroupsClusterPeer.java
```diff
@@ -144,13 +145,21 @@ public Address call()
                         udp.setMulticastPort(udp.getMulticastPort() + ((clusterName.hashCode() & 0x8fff_ffff) % portRangeLength));
                     }
 
+
+                    ProtocolStack stack = baseChannel.getProtocolStack();
+                    FORK fork = (FORK) stack.findProtocol(FORK.class);
+                    if (fork == null)
+                    {
+                        stack.insertProtocol(fork = new FORK(), ProtocolStack.ABOVE, FRAG2.class);
```
johnou: This shouldn't even be necessary, ForkChannel's forth argument create_fork_if_absent is set to true which does exactly the same thing inside the constructor.. https://github.com/belaban/JGroups/blob/master/src/org/jgroups/fork/ForkChannel.java

johnou: Now I see why you added it, must be a regression in JGroups..

Caused by: java.lang.NullPointerException
	at org.jgroups.protocols.FORK.createForkStack(FORK.java:199)
	at org.jgroups.fork.ForkChannel.<init>(ForkChannel.java:79)
	at com.ea.orbit.actors.cluster.JGroupsClusterPeer$1.call(JGroupsClusterPeer.java:148)
	... 7 more

johnou: https://github.com/belaban/JGroups/pull/241

PR #103, opened 2015-10-16
Allow setting of sticky headers.

johnou:
Motivation:

Currently the only way to set sticky headers is to use Orbit Container, it should be possible to set them when setting up Stage too.

Modification:

Add setter for execution and allow adding sticky headers to Execution.

Result:

Sticky headers can be set programmatically.

### DIFF
--- a/actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
+++ b/actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
@@ -56,8 +56,11 @@ import javax.inject.Singleton;
 
 import java.time.Clock;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
```diff
@@ -264,7 +370,7 @@ public void setMode(final StageMode mode)
         {
             if (container != null)
             {
-                if (clusterPeer == null && !container.getClasses().stream().filter(ClusterPeer.class::isAssignableFrom).findAny().isPresent())
```
johnou: null check is done above.

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
```diff
@@ -126,6 +134,104 @@
...
+        private String nodeName;
+        private StageMode mode = StageMode.HOST;
+        private int executionPoolSize = DEFAULT_EXECUTION_POOL_SIZE;
+
+        private Messaging messaging;
+
+        private List<ActorExtension> extensions = new ArrayList<>();
+        private Set<String> stickyHeaders = new HashSet<>();
+
+        public Builder clock(Clock clock) {
```
johnou: Unsure of what your naming strategy would be here clock vs setClock.

actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
```diff
@@ -493,6 +492,10 @@ public Activation(final ReferenceEntry entry, final Object instance)
         }
     }
 
+    public void addStickyHeaders(Collection<String> stickyHeaders)
```
johnou: No setter so developer doesn't accidently remove the transaction id sticky header.

DanielSperry:
I'd like to have more builders like this :)

PR #105, opened 2015-10-28
Adding Consistent Hash functionality to Hosting

OwenBorstad:
Can be used for load balancing and assigning work to specific Nodes.


DanielSperry:
@OwenBorstad, Thanks for the contribution!

PR #106, opened 2015-11-02
Prefer Local Placement

JoeHegarty:
This change wraps up a refactor of the general placement code to make it a little tider, it also includes the PreferLocalPlacement feature (as implemented in Orleans) with an additional percentile option for developers.


DanielSperry:
Looking good. I'll add some test cases for that later.

PR #107, opened 2015-11-02
Detailed TimeoutException message.

johnou:
In some cases we may want to know what actor calls are timing out.

### DIFF
--- a/actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
+++ b/actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
@@ -1240,7 +1240,7 @@ public class Execution implements Runtime
         return (T) reference;
     }
 
-    public Task<?> sendMessage(Addressable toReference, boolean oneWay, final int methodId, final Map<Object, Object> headers, final Object[] params)
+    public Task<?> sendMessage(Addressable toReference, Method m, boolean oneWay, final int methodId, final Map<Object, Object> headers, final Object[] params)
     {
         if (logger.isDebugEnabled())
         {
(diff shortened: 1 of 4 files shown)


actors/stage/src/main/java/com/ea/orbit/actors/runtime/MessageDefinitions.java
```diff
@@ -40,4 +40,5 @@
     public static final Byte INTERFACE_ID = 0;
     public static final Byte METHOD_ID = 1;
     public static final Byte OBJECT_ID = 2;
+    public static final Byte METHOD_NAME_ID = 4;
```
johnou: not sure why these are not primitives?

DanielSperry:
Those are going away in the next version...
They are going out of the headers.

JoeHegarty:
Further to Daniel's comments, we're going to reject this as it won't be neccesary in 0.6.x.
Please let us know if this continues to be a problem beyond then.

actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
```diff
@@ -1275,6 +1275,7 @@ private Throwable toSerializationSafeException(final Throwable notSerializable,
                 .withHeader(MessageDefinitions.INTERFACE_ID, actorReference._interfaceId())
                 .withHeader(MessageDefinitions.METHOD_ID, methodId)
                 .withHeader(MessageDefinitions.OBJECT_ID, ActorReference.getId(actorReference))
+                .withHeader(MessageDefinitions.METHOD_NAME_ID, m.getDeclaringClass().getName() + '.' + m.getName())
```
DanielSperry: We will try to avoid doing this.

PR #108, opened 2015-11-15
Added JAX-RS/JSON support to Orbit Web

JoeHegarty:
This change means that JAX-RS will work out of the box and have support for JSON.
I added a default healthcheck API as an example.
Finally, I tweaked the Embedded HTTP server so Orbit Container is also injectable via JAX-RS/Jersey/HK2.

### DIFF
--- a/web/pom.xml
+++ b/web/pom.xml
@@ -100,8 +100,14 @@
         </dependency>
 
         <dependency>
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
+            <artifactId>jackson-jaxrs-json-provider</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.glassfish.jersey.media</groupId>
-            <artifactId>jersey-media-moxy</artifactId>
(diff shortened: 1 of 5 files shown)


web/pom.xml
```diff
@@ -100,6 +100,18 @@
         </dependency>
 
         <dependency>
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
```
DanielSperry: I take that not having these dependencies where the only thing preventing the json based requests from working, right?

JoeHegarty: Pretty much. The ability to inject container is fairly important/useful though

DanielSperry: Cool. We will later need to find a way to return Task<...> from the  rest methods.

JoeHegarty: Yeh. I have some ideas on how to do a REST server an and an actor based impl. They are next on my list

quilombodigital: If you need help on this drop me a line. I dont know if you remember,  but some time ago I did a POC on this. https://github.com/quilombodigital/orbit/commits/bridge

PR #109, opened 2015-11-20
initial web socket support

DanielSperry:
Still unstable and missing some features.
Refactoring internal classes to make it easier to reuse code and to modify the protocol stack.


DanielSperry:
The central change is organizing the internal message handling like this:

diagram

That stack can be extended with plugins. And the components are reusable.

The rest of the change is reusing those components to build client connection stacks (with websockets for example).

PR #113, opened 2015-11-29
Allow orbit web JAX-RS APIs to return Tasks

JoeHegarty:
This is a very basic first pass to allow orbit web to return tasks in JAX-RS APIs.
Currently it blocks, this is required to correctly handle exceptions. In future I'd like to use code generation to avoid this, but I'd like to get the general programming model at least working so people can work against it.


web/pom.xml
```diff
@@ -95,7 +95,7 @@
 
         <dependency>
             <groupId>org.glassfish.jersey.containers</groupId>
-            <artifactId>jersey-container-servlet-core</artifactId>
+            <artifactId>jersey-container-servlet</artifactId>
```
JoeHegarty: This makes the servlet 3 stuff available

web/src/main/java/com/ea/orbit/web/provider/TaskProvider.java
```diff
@@ -0,0 +1,79 @@
...
+            Object result = entity;
+
+            try
+            {
+                result = ((Task<?>)entity).join();
+            }
+            catch(CompletionException e)
+            {
+                // We need to get the actual exception and rethrow it
+                throw (RuntimeException) e.getCause();
```
JoeHegarty: This is so the actual thrown exception is the one which is passed to Jersey.

web/src/main/java/com/ea/orbit/web/provider/TaskProvider.java
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            // Set generic type to null
+            // do we need to do anything special for wrapped types?
+            context.setGenericType(null);
```
JoeHegarty: I may need to do some work here if you have a nested generic type like Task<SomeGeneric<String>> or similar, going to test this.

JoeHegarty: Seems this actually works just fine, I've added a test covering it

web/src/main/java/com/ea/orbit/web/provider/TaskProvider.java
```diff
@@ -0,0 +1,74 @@
...
+            {
+                result = ((Task<?>)entity).join();
+            }
+            catch(CompletionException e)
+            {
+                // We need to get the actual exception and rethrow it
+                throw (RuntimeException) ((e.getCause() != null) ? e.getCause() : e);
+            }
+
+            context.setGenericType(null);
```
DanielSperry: We need to investigate the consequences of setGenericType(null)
I'd be more inclined to:
```
                        context.setGenericType(entityType == null || GenericTypeReflector.isSuperType(entityType, e.getClass())
                                // the actual entity class is narrower than task type parameter
                                // or the task type parameter is undefined or too generic
                                ? e.getClass()
                                // the entity type contains less information than the type parameter
                                // happens for instance if:
                                //    Task<List<SomeDto>> someMethod() { return Task.from(new ArrayList<>(value)) };
                                // where the entity class erases the type parameter
                                : entityType);
```

PR #118, opened 2015-12-02
Re-implementing Stage/Messaging Metrics as Pipeline Extension

FieldFlux:
* I've added back localActorCount back via Stage and execution since I didn't see an easy way to expose this information from an extension.

* Created a new Pipeline Extension for measuring Message Rates. It logs number of messages sent and received from a node along with rates of each message type in both directions. (Allows for monitoring error rates etc) 

* Fixed an issue in automatic metric registration where inherited metrics were not being registered.

### DIFF
--- a/actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsPipelineExtension.java
+++ b/actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsPipelineExtension.java
@@ -108,7 +108,9 @@ public class MetricsPipelineExtension extends NamedPipelineExtension
             Meter messageMeter = receiveMeters.get(messageType);
 
             if (messageMeter != null)
+            {
                 messageMeter.mark();
+            }
         }
 
         ctx.fireRead(message);
@@ -126,14 +128,16 @@ public class MetricsPipelineExtension extends NamedPipelineExtension
             Meter messageMeter = sendMeters.get(messageType);
 
             if (messageMeter != null)
+            {
                 messageMeter.mark();
+            }
         }
 
-        return super.write(ctx, message);
+        return ctx.write(message);
     }
 
     @ExportMetric(name="messagesReceived")
-    public int getMessagesRecievedCount()
+    public int getMessagesReceivedCount()
     {
         return messagesReceived.intValue();
     }
--- a/actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsPipelineExtension.java
+++ b/actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsPipelineExtension.java
@@ -1,0 +1,146 @@
+/*
+ Copyright (C) 2015 Electronic Arts Inc.  All rights reserved.
+
+ Redistribution and use in source and binary forms, with or without
+ modification, are permitted provided that the following conditions
+ are met:
+
+ 1.  Redistributions of source code must retain the above copyright
+     notice, this list of conditions and the following disclaimer.
+ 2.  Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
+ 3.  Neither the name of Electronic Arts, Inc. ("EA") nor the names of
+     its contributors may be used to endorse or promote products derived
+     from this software without specific prior written permission.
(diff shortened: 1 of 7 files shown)


JoeHegarty:
This looks good. I'll let @DanielSperry take a look but I'm happy to merge it as long as he is.

actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsPipelineExtension.java
```diff
@@ -0,0 +1,146 @@
...
+
+            if (messageMeter != null)
+                messageMeter.mark();
+        }
+
+        return super.write(ctx, message);
+    }
+
+    @ExportMetric(name="messagesReceived")
+    public int getMessagesRecievedCount()
```
DanielSperry: typo?

actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsPipelineExtension.java
```diff
@@ -0,0 +1,146 @@
...
+
+            Message msg = (Message) message;
+            final int messageType = msg.getMessageType();
+            Meter messageMeter = sendMeters.get(messageType);
+
+            if (messageMeter != null)
+                messageMeter.mark();
+        }
+
+        return super.write(ctx, message);
```
DanielSperry: use return ctx.write(message), looks better in the stack traces.

actors/stage/src/main/java/com/ea/orbit/actors/Stage.java
```diff
@@ -1225,4 +1226,16 @@ public String toString()
...
                 '}';
     }
+
+    @ExportMetric(name="localActorCount")
+    public int getLocalActorCount()
+    {
+        int value = 0;
+        if (execution != null)
+        {
+            value = execution.getObjectCount();
```
DanielSperry: actually the stage owns the `objects`, execution only has the pointer

actors/stage/src/main/java/com/ea/orbit/actors/runtime/Execution.java
```diff
@@ -164,4 +164,9 @@ public void setObjects(final LocalObjects objects)
         this.objects = objects;
     }
 
+    public int getObjectCount()
```
DanielSperry: not necessary. stage has the `objects`.

utils/metrics/metrics/src/main/java/com/ea/orbit/metrics/MetricsManager.java
```diff
@@ -256,7 +256,7 @@ private Metric buildGaugeForField(Field field, Object obj)
 
         if (obj != null)
         {
-            for (Field field : obj.getClass().getDeclaredFields())
+            for (Field field : obj.getClass().getFields())
```
DanielSperry: is this ok? it won't get private, default, and protected fields

DanielSperry:
pretty good.

I've pointed some minor things. 
@FieldFlux, If you don't have the time to fix them, we can merge tomorrow and fix them later.

johnou:
It would be nice if you could expose the stage metrics another way eg. via the pipeline. I do not use the orbit metrics package so when compiling there is always warnings about ExportMetric missing.

actors/extensions/metrics/src/main/java/com/ea/orbit/actors/extensions/metrics/MetricsPipelineExtension.java
```diff
@@ -0,0 +1,146 @@
...
+    {
+        if (message instanceof Message)
+        {
+            messagesReceived.increment();
+
+            Message msg = (Message) message;
+            final int messageType = msg.getMessageType();
+            Meter messageMeter = receiveMeters.get(messageType);
+
+            if (messageMeter != null)
```
johnou: missing braces?

utils/metrics/metrics/src/main/java/com/ea/orbit/metrics/MetricsManager.java
FieldFlux: Thats ok for now. We didn't register inaccessible fields anyway. I plan on looking at this for a future change.

DanielSperry:
Thanks @FieldFlux !

PR #122, opened 2015-12-12
Utility class which provides cache for expensive annotation lookups.

johnou:


DanielSperry:
@johnou, thanks for the contribution!

PR #123, opened 2015-12-14
Variable substitution for the Orbit config file

Fiurgeist:
Usage:
- orbit.http.port: ${http.port} -> name of the system property
- orbit.http.port: ${http.port:1337} -> name of the system property and a default value separated by a colon ':'

---

I need the possibility to set Orbit config values over system properties, especially for ports, because in my set-up the application becomes random ports provided on start-up, so I can't have them hard-coded in the config file. I used an approach you can find for other config files (e.g. for JGroups or Hazelcast), looking for ${...} and checking if there is a system property with that name.


DanielSperry:
This look ok to me. 
@JoeHegarty, how about you?

JoeHegarty:
The build is failing currently, so we can't merge until it's green.

JoeHegarty:
Ok, looks like it';s passing now. Intermittent failure.

JoeHegarty:
Merged. Thanks for the contribution!

PR #125, opened 2016-01-08
Support for marking immutable objects to avoid cloning.

johnou:


DanielSperry:
It's all or nothing, right?

If a single parameter is not immutable everything gets cloned together:
to avoid problems with graphs:  (immutable p1, mutable p2={a=p1})

johnou:
Correct, but on second thought it might make more sense to extract all this immutable logic into a helper that the  cloners can use wdyt?

johnou:
Yep or cloning would get pretty messy in my opinion.

DanielSperry:
cool let's merge it.

PR #126, opened 2016-01-11
Observer enhancements.

johnou:
Introduced new ObserverMapManager.
Less object creation when invoking notifyObservers.

### DIFF
--- a/actors/core/src/main/java/com/ea/orbit/actors/ObserverManager.java
+++ b/actors/core/src/main/java/com/ea/orbit/actors/ObserverManager.java
@@ -102,7 +102,6 @@ public class ObserverManager<T extends ActorObserver> implements Serializable
      * It's not recommended to wait on this task since it might take a while to finish the
      * cleanup process and ObserverManager is thread safe.
      */
-    @SuppressWarnings("unchecked")
     public Task<?> cleanup() {
         // TODO: replace ping for a single batch call for each node containing observers from this list.
         // TODO: add a observer validation function to the runtime. ActorRuntime
--- a/actors/core/src/main/java/com/ea/orbit/actors/ObserverManager.java
+++ b/actors/core/src/main/java/com/ea/orbit/actors/ObserverManager.java
@@ -28,8 +28,6 @@
 
 package com.ea.orbit.actors;
 
-import com.ea.orbit.actors.cluster.NodeAddress;
-import com.ea.orbit.actors.runtime.ActorRuntime;
 import com.ea.orbit.actors.runtime.RemoteReference;
 import com.ea.orbit.concurrent.ConcurrentHashSet;
 import com.ea.orbit.concurrent.Task;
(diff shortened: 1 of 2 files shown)


johnou:
@DanielSperry I had a look at implementing the batch ping TODO but couldn't come up with an elegant way to query each node if their observer is alive..

```java
Task<Collection<? extends ActorObserver>> batchPing(Collection<? extends ActorObserver> references);
```

actors/core/src/main/java/com/ea/orbit/actors/ObserverManager.java
```diff
@@ -129,22 +144,26 @@ public boolean addObserver(final T observer)
...
+        {
             try
             {
-                callable.accept(o);
+                callable.accept(observer);
             }
             catch (final Exception ex)
             {
-                fail.add(o);
+                if (fail == null) {
```
johnou: I considered invoking remove directly here but looking at the source for remove and removeAll I suspect the later takes less CPU time with large collections.

JoeHegarty:
I'd like @DanielSperry to chime in on this one before I merge it.

DanielSperry:
Hey guys, I will be able to look into this next Thursday.

actors/core/src/main/java/com/ea/orbit/actors/ObserverManager.java
```diff
@@ -98,24 +102,35 @@ public boolean addObserver(final T observer)
...
+                .map(observer ->
+                        (observer.ping()).handle((Object result, Throwable ex) -> {
+                            if (ex != null)
                             {
-                                // beware: this might run in parallel with other calls the actor.
-                                // this shouldn't be a problem.
-                                observers.remove(o);
-                            }
-                        }));
+                                return checkAlive(observer).thenAccept(alive -> {
```
DanielSperry: We should remove the observers when they no longer exist in the remote node.
This change will let old zombie observers exist until their original node restarts.

I'll take this change to another branch and add the method @johnou cited:
a call to the original node confirming that the observer is alive.
this will remove the ping.

The new flow will be:
cleanup:
- is the original node still up? no: remove observer
- is the observer alive in the node? no: remove observer.
- keep observer

DanielSperry:
We can't merge this ATM. It will change the cleanup behaviour.

actors/core/src/main/java/com/ea/orbit/actors/ObserverManager.java
johnou: not so sure about "is the original node still up? no: remove observer" what if a node drops from the cluster and rejoins in a short period?

johnou:
Removed the cleanup behaviour changes.

JoeHegarty:
Do we want to merge this now @DanielSperry ?

DanielSperry:
It looks good to merge!

PR #127, opened 2016-01-12
Return getClassById instead of null.

johnou:


johnou:
com.ea.orbit.exception.UncheckedException: Class not found, id: -1055249813

	at com.ea.orbit.actors.runtime.DefaultDescriptorFactory.ref(DefaultDescriptorFactory.java:223)

PR #128, opened 2016-01-12
Detailed logging for missing counterpart.

johnou:


actors/stage/src/main/java/com/ea/orbit/actors/runtime/Messaging.java
```diff
@@ -304,9 +332,10 @@ public Task write(final HandlerContext ctx, final Object msg) throws Exception
         RemoteReference<?> actorReference = (RemoteReference<?>) toReference;
         NodeAddress toNode = invocation.getToNode();
 
-        final LinkedHashMap<Object, Object> actualHeaders = new LinkedHashMap<>();
+        LinkedHashMap<Object, Object> actualHeaders = null;
         if (invocation.getHeaders() != null)
         {
+            actualHeaders = new LinkedHashMap<>();
```
johnou: Less object creation, should be safe as withHeader has a null check which will instantiate it within Message pojo if necessary.

PR #130, opened 2016-01-26
Stream message dispatcher for invoking requests to annotated methods.

johnou:


JoeHegarty:
Seems reasonable to me, @DanielSperry ?

actors/test/actor-tests/src/test/java/com/ea/orbit/actors/test/streams/StreamMessageDispatcherTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @SuppressWarnings("Duplicates")
+    @Test(timeout = 30_000L)
+    public void testStreamMessageDispatcher()
+    {
+        createStage();
+        StreamMessageDispatcher messageDispatcher = new StreamMessageDispatcher(Collections.singleton(this));
+        try
+        {
+            AsyncStream<Object> test = AsyncStream.getStream(Object.class, "test");
+            test.subscribe((d, t) -> {
```
johnou: We could provide some wrapper for dealing with subscription in the constructor or some builder / factory?

DanielSperry:
It's an util class, right?
looks good.

Thanks, @johnou

johnou:
Yep, use case is that I have an actor which will receive several different message types over a stream and having a bunch of instance of checks didn't seem desirable.

PR #135, opened 2016-02-09
Actor method annotated with CacheResponse which returns null will throw a hidden NPE.

johnou:
Actor method annotated with CacheResponse which returns null will throw a hidden NPE.

Motivation:

If an actor method annotated with CacheResponse returns null it will throw a hidden NPE.

Modifications:

Fix possible NPE, add unit test and support for enums.

Result:

Actor method annotated with CacheResponse which returns null will no longer throw an NPE.


JoeHegarty:
Would you be able to resubmit this PR with the new project structure?

johnou:
Yep no worries.

PR #137, opened 2016-02-16
Fixed reminder bug and introduced NoDeactivate annotation.

JoeHegarty:
This fixes a reported issue of reminders not always firing. It also introduces the NoDeactivate annotation which helps fix that issue.

### DIFF
--- a/actors/core/src/main/java/com/ea/orbit/actors/annotation/NeverDeactivate.java
+++ b/actors/core/src/main/java/com/ea/orbit/actors/annotation/NeverDeactivate.java
@@ -35,6 +35,6 @@ import java.lang.annotation.Target;
 
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
-public @interface NoDeactivate
+public @interface NeverDeactivate
 {
 }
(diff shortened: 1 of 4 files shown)


johnou:
Not a super fan of the annotation name but I cannot think of something else at the moment..

FieldFlux:
Perhaps the annotation can be called AlwaysActive? or AlwaysActivated? I prefer Positive Logic when naming things :)

JoeHegarty:
My only concern with "AlwaysActivated" is it suggests to me that the framework is going to activate it for you, but that's not the case. You still need to trigger the initial activation.

FieldFlux:
Hmm, yeah, I hadn't thought of that. Maybe NeverDeactivate instead of NoDeactivate?

JoeHegarty:
NeverDeactivate works for me.

DanielSperry:
What happens on node shutdown/crash for NeverDeactivate actors?

JoeHegarty:
You'll need to activate it on another node. The framework won't do anything special in that case.

JoeHegarty:
We could name it more like PreferLocalPlacement.
Make it PreferPerpetual or PreferPermanent or similar?

PR #140, opened 2016-03-01
Allow actor implementations to hint to the system when the actor can …

johnou:
…be deactivated.

Motivation:

Last access TTL based deactivation is not suitable in all scenarios.

Modifications:

Add new method to AbstractActor which implementations can override to hint to the system when the actor can be deactivated.

Result:

Better lifetime management of actors.

### DIFF
--- a/actors/runtime/src/main/java/com/ea/orbit/actors/runtime/ActorEntry.java
+++ b/actors/runtime/src/main/java/com/ea/orbit/actors/runtime/ActorEntry.java
@@ -164,13 +164,34 @@ public class ActorEntry<T extends AbstractActor> extends ActorBaseEntry<T>
             {
                 return Task.done();
             }
-            return executionSerializer.offerJob(key, () -> doDeactivate(), 1000);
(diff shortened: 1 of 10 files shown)


JoeHegarty:
Thanks @johnou. Going to take me a while to review this, I'll take a look this afternoon.

johnou:
I am not overly keen on the performance downside of this pull request, if I have time later I will investigate implementing DeactivateOnIdle à la Orleans as suggested by @blaager.

actors/runtime/src/main/java/com/ea/orbit/actors/runtime/ActorEntry.java
```diff
@@ -164,7 +164,7 @@ public T getObject()
             {
                 return Task.done();
             }
-            return executionSerializer.offerJob(key, () -> doDeactivate(), 1000);
```
johnou: Still might be a good idea to increase this regardless.

PR #141, opened 2016-03-16
Adding support for reentrant actors

johnou:
Motivation:

Adding support for reentrant actors greatly improves performance for actors which return chained tasks.

Modifications:

Orbit execution serializer now supports interleaving requests.

Result:

Support for reentrant actors.


johnou:
Unfortunately we did see a slight drop in performance with the SingleNodeBenchmark, but I think the performance gained from interleaving request support is worth the trade.

Before:
SingleNodeBenchmark.requestThroughput            thrpt  100  2072540.407 ± 22749.407  ops/s

After:
SingleNodeBenchmark.requestThroughput            thrpt  100  1917053.794 ± 24278.528  ops/s

JoeHegarty:
Thanks a lot for this!

PR #143, opened 2016-03-30
Fix test.

johnou:
Added missing Reentrant annotation.


JoeHegarty:
I'll take this fix and we'll keep an eye out for the issue reoccuring.

Thanks! Joe

PR #144, opened 2016-04-02
Replace Guava with Caffeine.

johnou:
Caffeine vs Guava (old benchmarks below, see commit for updated benchmarks)

Caffeine 

Benchmark                                         Mode  Cnt        Score       Error  Units
SingleNodeBenchmark.requestThroughput            thrpt  100  3013608.693 ± 29675.239  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt  100        1.242 ±     0.007  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt  100        4.279 ±     0.060  us/op

Benchmark                                         Mode  Cnt        Score       Error  Units
SingleNodeBenchmark.requestThroughput            thrpt  100  3085224.027 ± 36085.998  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt  100        1.269 ±     0.006  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt  100        4.142 ±     0.074  us/op

Guava

Benchmark                                         Mode  Cnt        Score       Error  Units
SingleNodeBenchmark.requestThroughput            thrpt  100  3152836.327 ± 35103.067  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt  100        1.239 ±     0.010  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt  100        4.081 ±     0.078  us/op

### DIFF
--- a/commons/src/main/java/cloud/orbit/concurrent/Task.java
+++ b/commons/src/main/java/cloud/orbit/concurrent/Task.java
@@ -66,8 +66,7 @@ public class Task<T> extends CompletableFuture<T>
     private static Void NIL = null;
 
     private static Executor commonPool = ExecutorUtils.newScalingThreadPool(100);
-    private static final ScheduledExecutorService schedulerExecutor =
-            Executors.newScheduledThreadPool(10);
+    private static ScheduledExecutorService schedulerExecutor = Executors.newScheduledThreadPool(10);
 
     // TODO: make all callbacks async by default and using the current executor
     // what "current executor' means will have to be defined.
--- a/actors/infinispan-cluster/src/main/resources/conf/jgroups.xml
+++ b/actors/infinispan-cluster/src/main/resources/conf/jgroups.xml
@@ -26,7 +26,6 @@
   (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
   THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   -->
-
 <!--
   Default stack using IP multicasting. It is similar to the "udp"
   stack in stacks.xml, but doesn't use streaming state transfer and flushing
--- a/actors/runtime/pom.xml
+++ b/actors/runtime/pom.xml
@@ -94,9 +94,9 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
         </dependency>
 
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>18.0</version>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>2.2.6</version>
         </dependency>
 
         <dependency>
--- a/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
@@ -115,6 +115,7 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ForkJoinTask;
 import java.util.concurrent.TimeUnit;
@@ -450,7 +451,6 @@ public class Stage implements Startable, ActorRuntime
             finder.start().join();
         }
 
-
         cacheManager = new ResponseCaching();
 
         hosting.setNodeType(mode == StageMode.HOST ? NodeCapabilities.NodeTypeEnum.SERVER : NodeCapabilities.NodeTypeEnum.CLIENT);
@@ -684,7 +684,9 @@ public class Stage implements Startable, ActorRuntime
                 }
 
                 // Skip this actor if it is marked NeverDeactivate
-                if(RemoteReference.getInterfaceClass(actorEntry.getRemoteReference()).isAnnotationPresent(NeverDeactivate.class)) continue;
+                if (RemoteReference.getInterfaceClass(actorEntry.getRemoteReference()).isAnnotationPresent(NeverDeactivate.class)) {
+                    continue;
+                }
 
                 if (clock().millis() - actorEntry.getLastAccess() > maxAge)
                 {
--- a/actors/runtime/src/main/java/cloud/orbit/actors/concurrent/WaitFreeMultiExecutionSerializer.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/concurrent/WaitFreeMultiExecutionSerializer.java
@@ -28,17 +28,17 @@
 
 package cloud.orbit.actors.concurrent;
 
-import cloud.orbit.actors.runtime.InternalUtils;
-import cloud.orbit.concurrent.ExecutorUtils;
-import cloud.orbit.concurrent.Task;
-import cloud.orbit.exception.UncheckedException;
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+
+import cloud.orbit.actors.runtime.InternalUtils;
+import cloud.orbit.concurrent.ExecutorUtils;
+import cloud.orbit.concurrent.Task;
+
 import java.util.List;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
@@ -52,11 +52,12 @@ import java.util.function.Supplier;
 public class WaitFreeMultiExecutionSerializer<T> implements MultiExecutionSerializer<T>
 {
     private static final Logger logger = LoggerFactory.getLogger(WaitFreeMultiExecutionSerializer.class);
+
     private final ExecutorService executorService;
 
     // while running, the WaitFreeExecutionSerializer is held alive by references from the executorService
     // and from anyone holding the promises (Tasks) it returns.
-    private Cache<T, WaitFreeExecutionSerializer> serializers = CacheBuilder.newBuilder().weakValues().build();
+    private final Cache<T, WaitFreeExecutionSerializer> serializers = Caffeine.newBuilder().weakValues().build();
 
     public WaitFreeMultiExecutionSerializer()
     {
@@ -73,14 +74,7 @@ public class WaitFreeMultiExecutionSerializer<T> implements MultiExecutionSerial
         final WaitFreeExecutionSerializer serializer = serializers.getIfPresent(key);
         if (serializer == null)
         {
-            try
-            {
-                return serializers.get(key, () -> new WaitFreeExecutionSerializer(executorService, key));
-            }
-            catch (ExecutionException e)
-            {
-                throw new UncheckedException(e);
-            }
+            return serializers.get(key, o -> new WaitFreeExecutionSerializer(executorService, key));
         }
         return serializer;
     }
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/ActorEntry.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/ActorEntry.java
@@ -164,7 +164,7 @@ public class ActorEntry<T extends AbstractActor> extends ActorBaseEntry<T>
             {
                 return Task.done();
             }
-            return executionSerializer.offerJob(key, () -> doDeactivate(), 1000);
+            return executionSerializer.offerJob(key, () -> doDeactivate(), 10000);
         }
         catch (Throwable ex)
         {
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/LocalObjects.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/LocalObjects.java
@@ -28,6 +28,10 @@
 
 package cloud.orbit.actors.runtime;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.RemovalCause;
+
 import cloud.orbit.actors.Actor;
 import cloud.orbit.actors.ActorObserver;
 import cloud.orbit.actors.cluster.NodeAddress;
@@ -35,25 +39,18 @@ import cloud.orbit.actors.transactions.IdUtils;
 import cloud.orbit.concurrent.Task;
 import cloud.orbit.concurrent.TaskFunction;
 import cloud.orbit.exception.NotImplementedException;
-import cloud.orbit.exception.UncheckedException;
-
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.RemovalCause;
-import com.google.common.cache.RemovalNotification;
 
 import java.util.ConcurrentModificationException;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.ExecutionException;
 import java.util.stream.Stream;
 
 public class LocalObjects
 {
-    private ConcurrentMap<Object, LocalObjectEntry> localObjects = new ConcurrentHashMap<>();
-    private Cache<Object, LocalObjectEntry> objectMap = CacheBuilder.newBuilder().weakKeys()
+    private final ConcurrentMap<Object, LocalObjectEntry> localObjects = new ConcurrentHashMap<>();
+    private final Cache<Object, LocalObjectEntry> objectMap = Caffeine.newBuilder().weakKeys()
             .removalListener(this::onRemoval).build();
 
 
@@ -245,24 +242,17 @@ public class LocalObjects
             throw new IllegalArgumentException("Object clashes with a pre-existing object: " + reference);
         }
         LocalObjectEntry localObject = createLocalObjectEntry(reference, object);
-        try
+        if (object != null)
         {
-            if (object != null)
+            final LocalObjectEntry other = objectMap.get(object, o -> localObject);
+            if (localObject != other)
             {
-                final LocalObjectEntry other = objectMap.get(object, () -> localObject);
-                if (localObject != other)
+                if (!Objects.equals(reference, other.getRemoteReference()))
                 {
-                    if (!Objects.equals(reference, other.getRemoteReference()))
-                    {
-                        throw new ConcurrentModificationException();
-                    }
-                    return (RemoteReference) other.getRemoteReference();
+                    throw new ConcurrentModificationException();
                 }
+                return (RemoteReference) other.getRemoteReference();
             }
-        }
-        catch (ExecutionException e)
-        {
-            throw new UncheckedException(e);
         }
         final LocalObjectEntry previous = localObjects.putIfAbsent(reference, localObject);
         if (previous != null && localObject != previous)
@@ -312,14 +302,13 @@ public class LocalObjects
         return localObject;
     }
 
-    private void onRemoval(final RemovalNotification<Object, LocalObjectEntry> objectObjectRemovalNotification)
+    private void onRemoval(final Object key, final LocalObjectEntry value, final RemovalCause removalCause)
     {
-        if (objectObjectRemovalNotification.getCause() == RemovalCause.COLLECTED)
+        if (removalCause == RemovalCause.COLLECTED)
         {
-            final LocalObjectEntry objectEntry = objectObjectRemovalNotification.getValue();
-            if (objectEntry != null && objectEntry.getRemoteReference() != null)
+            if (value != null && value.getRemoteReference() != null)
             {
-                localObjects.remove(objectEntry.getRemoteReference());
+                localObjects.remove(value.getRemoteReference());
             }
         }
     }
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Messaging.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Messaging.java
@@ -67,7 +67,7 @@ public class Messaging extends HandlerAdapter implements Startable
 
     private final LongAdder networkMessagesReceived = new LongAdder();
     private final LongAdder responsesReceived = new LongAdder();
-    private static Timer timer = new Timer("Messaging timer");
+    private static Timer timer = new Timer("Messaging timer", true);
     private BasicRuntime runtime;
 
 
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/ResponseCaching.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/ResponseCaching.java
@@ -42,9 +42,9 @@ import cloud.orbit.exception.UncheckedException;
 import cloud.orbit.tuples.Pair;
 import cloud.orbit.util.AnnotationCache;
 
-import com.google.common.base.Ticker;
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.Ticker;
 
 import java.io.IOException;
 import java.io.OutputStream;
@@ -53,6 +53,7 @@ import java.math.BigInteger;
 import java.security.DigestOutputStream;
 import java.security.MessageDigest;
 import java.util.List;
+import java.util.concurrent.Executor;
 import java.util.stream.Collectors;
 
 public class ResponseCaching
@@ -60,6 +61,7 @@ public class ResponseCaching
         implements ExecutionCacheFlushObserver
 {
     private static Ticker defaultCacheTicker = null;
+    private static Executor cacheExecutor = null;
     private MessageSerializer messageSerializer;
     private BasicRuntime runtime;
     private final AnnotationCache<CacheResponse> cacheResponseCache = new AnnotationCache<>(CacheResponse.class);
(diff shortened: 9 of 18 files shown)


johnou:
@ben-manes excellent work with Caffeine, basically a drop in replacement for the Orbit project. Unfortunately I am seeing a bit of a performance drop when migrating from Guava 18.0 to Caffeine 2.2.3, any tips?

johnou:
Note to self, check Failed tests:   testCacheDuration(cloud.orbit.actors.test.CacheResponseTest): expected:<2> but was:<3>

johnou:
I suspect testCacheDuration is failing on CI because Guava amortizes maintenance work on the calling thread during read and write operations whereas Caffeine maintenance is asynchronous.

ben-manes:
That's surprising, I'll need to run those benchmarks and investigate. The delegation to a background thread is to reduce response times, but in a benchmark is a penalty due to the extra context switches.

My initial suspicion is to compare their performance with weak keys, as it's not a case I've benchmarked. Guava's forked the hash table to optimize for reference collection, whereas I decorated instead and optimized for maximum size. That requires a wrapper lookup key, so especially on a lot of misses the extra GC would be a penalty.

ben-manes:
Oh, and for tests the simplest is to set `Caffeine.executor(Runnable::run)` which is usually easy if you're using DI. If not then you're right using `cleanUp` is probably the simplest.

In the benchmark below the caches are built as `Caffeine.newBuilder().weakKeys().build()` and `CacheBuilder.newBuilder().concurrencyLevel(CONCURRENCY_LEVEL).weakKeys().build()`. In that scenario Guava wins on reads when not overwhelmed (contention on the per-segment counter). Since this is such a high penalty compared to a raw `ConcurrentHashMap`, it points to the lookup key wrapper as the culprit. The wrapper is created per lookup to compare by identity (identity hash and equals) which Guava can do directly in the hash table (via `Equivalence<Object> keyEquivalence`). I think the overhead would be smaller if I recycle the wrapper using a `ThreadLocal` instead, where the extra hashing will be a lower cost than the wrapper's GC penalty.

Can you quickly hack your changes to compare w/o weak keys enabled? That would be useful confirmation.

```
Benchmark                   (cacheType)   Mode  Cnt         Score          Error  Units
GetPutBenchmark.read_only            Caffeine  thrpt   10   78904420.319 ±  4338914.946  ops/s
GetPutBenchmark.read_only            Guava_4   thrpt   10   65837382.656 ±  4424105.297  ops/s
GetPutBenchmark.read_only            Guava_64  thrpt   10   85528167.301 ±  2885286.447  ops/s
GetPutBenchmark.read_only   ConcurrentHashMap  thrpt   10  259622836.318 ± 13136471.220  ops/s
GetPutBenchmark.readwrite            Caffeine  thrpt   10   72009974.220 ± 10445241.080  ops/s
GetPutBenchmark.readwrite             Guava_4  thrpt   10   46887322.041 ±  3901540.891  ops/s
GetPutBenchmark.readwrite            Guava_64  thrpt   10   71763490.732 ±  2384287.217  ops/s
GetPutBenchmark.readwrite   ConcurrentHashMap  thrpt   10  201025963.215 ±  7860052.150  ops/s
GetPutBenchmark.write_only           Caffeine  thrpt   10   29298544.080 ±   815351.349  ops/s
GetPutBenchmark.write_only            Guava_4  thrpt   10    4953373.921 ±   333589.144  ops/s
GetPutBenchmark.write_only           Guava_64  thrpt   10   19953776.079 ±   706751.768  ops/s
GetPutBenchmark.write_only  ConcurrentHashMap  thrpt   10   66842333.531 ±  5259936.811  ops/s
```

ben-manes:
Due to escape analysis the wrapper isn't allocated on a read, so that wasn't the issue.

The cause is because `ForkJoinTask.RunnableExecuteAction` is allocated on every submission to wrap the runnable, which performs the cache's maintenance. For `weakKeys` this results in millions in the profiler, with only thousands for `maximumSize`. That's probably because the task execution is much faster, so the submission rate is higher. The solution is to allocate the a `ForkJoinTask` instead of the runnable and the pool reuses it. I remember seeing this before and I don't know how I forgot to fix it.

```
Benchmark                 (cacheType) Mode   Cnt          Score         Error  Units
GetPutBenchmark.read_only   Caffeine  thrpt   10  149676722.976 ± 6005882.472  ops/s
GetPutBenchmark.readwrite   Caffeine  thrpt   10  104603504.149 ± 6881197.985  ops/s
GetPutBenchmark.write_only  Caffeine  thrpt   10   37331074.722 ± 1432268.090  ops/s
```

In the case of a write the performance drop is due to the the wrapper allocation, as escape analysis fails. I think this is because the method is too long, so the JIT gives up on that optimization. Unfortunately simply moving the call to another method doesn't help reliably, as most of the time the JIT will decide to inline the call. This method is a challenge to break apart and I'm not sure if its worth the effort. The time when the JIT failed to inline provided a 10M boost, which is lower than I expected (should be closer to 60M ops/s).

ben-manes:
Hi @johnou,

Can you retest using the `SNAPSHOT`? This contains the improvements described earlier. I haven't found a good way to speed up writes. That is still faster than Guava and should not be the source of a bottleneck.

ben-manes:
Oh, I spoke too soon. Looks like I was able to raise the writeOnly throughput to 50M ops/s by pre-calculating the identity hash code. This probably won't affect most common usages, though, because it was only re-calculated on a retrial due to concurrent writes to the same entry. A retry is common in the benchmarks due to using a Zipf distribution to emulate real-world access patterns.

johnou:
@ben-manes apologies for not getting back to you sooner, cat emergency. Impressive detailed analysis, I will try the latest SNAPSHOT now.

ben-manes:
Those things take priority, there's no rush on my side.

The write improvement might have been a fluke, as I can't reproduce. Benchmarking on a laptop is error prone...

ben-manes:
Ahh, I see now. It makes it easier for escape analysis to detect because the wrapper is a simple value object. If pulled into the retry loop the JIT might be smart and avoid the allocation. If outside then it sometimes is smart, sometimes allocates on the heap.

johnou:
2.2.4-SNAPSHOT is almost on par, I will run another couple of benchmarks to verify.

Benchmark                                         Mode  Cnt        Score       Error  Units
SingleNodeBenchmark.requestThroughput            thrpt  100  3102322.063 ± 36487.435  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt  100        1.239 ±     0.007  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt  100        4.228 ±     0.072  us/op

johnou:
2.2.4-SNAPSHOT without weak values.

Benchmark                               Mode  Cnt        Score       Error  Units
SingleNodeBenchmark.requestThroughput  thrpt  100  3051152.122 ± 58227.558  ops/s

ben-manes:
Hard to say. The cache might not be a bottleneck and the extra context switches are visible. You might try using `Caffeine.executor(Runnable::run)` to make Caffeine amortize on user threads. The main difference, though, is it delegates the removal notifications to the executor so might be less evenly spread out (Guava publishes to a queue, unlocks, and all requests help process it).

I tried running your benchmarks but it failed due to a connection issue (no server running?). The next step would be to profile Orbit and to benchmark with the other cache features that you're using (weakValues, expireAfterWrite) in case there's another oversight.

ben-manes:
It may also be that `cache.get(key, func)` doesn't get escaped analysis so there is the GC issue on that code path.

johnou:
Can you be more specific on what connection issue? SingleNodeBenchmark should run without any external services.

ben-manes:
I might have misread it.

There are warnings in the warm-up,
```
14:40:27.560 [ForkJoinPool.commonPool-worker-1] WARN  org.jgroups.protocols.UDP - failed to join /224.0.75.75:7500 on utun0: java.net.SocketException: Can't assign requested address
```

and it fails after 20 iterations with,
```
Iteration  20: 14:41:10.209 [cloud.orbit.actors.test.SingleNodeBenchmark.requestThroughput-jmh-worker-3] WARN  cloud.orbit.actors.net.DefaultHandlerContext - Unprocessed write STOPPING
2182641.664 ops/s
<JMH had finished, but forked VM did not exit, are there stray running threads? Waiting 24 seconds more...>
```

ben-manes:
It looks like there's some non-daemon executors?

```
Thread[pool-3-thread-5,5,main]
  at sun.misc.Unsafe.park(Native Method)
  at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
  at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
  at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:1088)
  at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:809)
  at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1067)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1127)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
  at java.lang.Thread.run(Thread.java:745)
```

johnou:
Yes there is, I should look into that. JMH kills the forks after 30 seconds so it slows down the entire benchmarking process. My colleague had the same issue as you, iirc he worked around it by disabling WIFI. I will change JGroups so you can run it without problems.

Below is benchmark with setting executor(Runnable::run)

Benchmark                               Mode  Cnt        Score       Error  Units
SingleNodeBenchmark.requestThroughput  thrpt  100  3051610.575 ± 43185.655  ops/s

ben-manes:
Oh the another reason might be due to the ticker, which calls `System.nanoTime()`. On Mac/Windows that can be quite slow, especially with contention, due to polling hardware. On Linux it is very fast as a kernel thread updates a snapshot value, which is what all 3 platforms do with `System.currentTimeMillis()`. If Caffeine is faster then more contention on the time occurs and there is a net loss.

Can you try using a custom ticker with millis or just returning back a constant?

Sorry that this is a game of debugging and not a smooth migration.

johnou:
Would you please try running SingleNodeBenchmark with -Djava.net.preferIPv4Stack=true? I am currently running benchmarks with .executor(Runnable::run).ticker(System::currentTimeMillis)

ben-manes:
Will do.

Expiration is definitely causing a lot of overhead.
```
# System.nanoTime()
Benchmark  (cacheType) Mode  Cnt         Score         Error  Units
read_only   Caffeine  thrpt   10  34119040.917 ± 1237210.222  ops/s
read_only      Guava  thrpt   10  25222603.741 ± 1743789.896  ops/s
readwrite   Caffeine  thrpt   10  11702791.308 ± 2444290.601  ops/s
readwrite      Guava  thrpt   10  25538939.002 ±  715190.238  ops/s
write_only  Caffeine  thrpt   10   2381556.496 ±   71545.187  ops/s
write_only     Guava  thrpt   10   8410739.672 ±  206273.140  ops/s

# System.currentTimeMillis()
Benchmark(cacheType)   Mode  Cnt         Score         Error  Units
read_only   Caffeine  thrpt   10  68226450.796 ± 4746074.829  ops/s
read_only      Guava  thrpt   10  49128190.595 ± 1542803.952  ops/s
readwrite   Caffeine  thrpt   10  21119924.971 ± 6150580.330  ops/s
readwrite      Guava  thrpt   10  45130393.723 ± 1425360.779  ops/s
write_only  Caffeine  thrpt   10   2401491.244 ±  112033.233  ops/s
write_only     Guava  thrpt   10   7841423.896 ±  455622.536  ops/s
```

ben-manes:
It looks like there's a regression in expiration and it will take me a while to figure out why it is so slow. Right now I see its much worse than expected, but I'm not sure yet why.

ben-manes:
Sorry this is going to be a long explanation and you can ignore it. But if interested I think it is worth explaining why this slow down is observed.  I don't this affects your benchmarks much, but is visible in mine.

---

The issue only occurs for `expireAfterWrite`, as `expireAfterAccess` is only slowed down by acquiring the time. The problem is a case of queuing theory, thread scheduling, and a very high write rate.

In Guava a write blocks the segment and performs the all of the maintenance work for it thus far. This causes other writes (1 / n where by default n=4) to deschedule and wait their turn, as the operation is fairly slow given how broad it is. Reads progress and don't have to do the maintenance work because the writes are taking care of that. This penalizes writes over reads, but that's desirable because reads are the common case and latency sensitive. The parallelism of writes (N segments) helps to some degree.

Caffeine avoids segments as it doesn't use any stripe locks nor is ConcurrentHashMap segmented anymore. Just like how reads are recorded to a buffer and replayed under a lock for maintenance work, writes are buffered as well. The difference being that while the read buffer is lossy and scheduling can be delayed, writes must be kept and the maintenance work is scheduled immediately. This lets writers proceed and a single background thread takes care of the maintenance. In a real system this reduces user-facing latency and there's headroom to do this processing async.

In a benchmark there is no application business logic between writes and instead a writer gets the full cpu time slice to keep writing into the cache. With many writers into the queue and a single thread draining it, the overall performance begins to drop given the GC and other affects. I saw this in my maximumSize write [stress tests](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/test/java/com/github/benmanes/caffeine/cache/Stresser.java) for inserts, but it wasn't an issue because a `Thread.yield()` to emulate the real world resolved the issue.

In most updates it bypasses the write buffer and treats it as a read, e.g. an LRU reorder is the same in either case. It is only on a weight change or `expireAfterWrite` is that not true. The latter is because we order expiration on a FIFO queue, so like an LRU we can reorder in O(1) time and expire by peeking at the head. Because reads only reorder by access and we enqueue the node to avoid GC, the write buffer is used. This results in the same worse-case scenario under an artificial load that causes a degredation.

The solution is then to take the same fuzzy logic that memcached and others use - ignore it if within a small time window (e.g. 1 second). That classic trick is most often to avoid lock contention on reads for an LRU, but here we can use it to avoid overwhelming the write queue. If we observe the write rate is too high and that updating the entry's write time won't have a visible effect, we can ignore the work.

With that change the bottleneck becomes obtaining the system time, which is a slow operation.

```
# Using System.nanoTime() [high resolution]
Benchmark (cacheType) Mode  Cnt         Score         Error  Units
read_only  Caffeine  thrpt   10  33300860.072 ± 2059302.969  ops/s
read_only     Guava  thrpt   10  25696096.709 ± 1216610.169  ops/s
readwrite  Caffeine  thrpt   10  29352638.708 ± 1067015.454  ops/s
readwrite     Guava  thrpt   10  25962527.483 ±  530785.662  ops/s
write_only Caffeine  thrpt   10  27523761.263 ±  550693.590  ops/s
write_only    Guava  thrpt   10   8182072.940 ±  142045.763  ops/s

# Using System.currentTimeMillis() [medium resolution]
Benchmark  (cacheType) Mode  Cnt    Score          Error     Units
read_only  Caffeine  thrpt   10  65645152.423 ± 7583198.354  ops/s
read_only     Guava  thrpt   10  46095795.915 ± 5639592.232  ops/s
readwrite  Caffeine  thrpt   10  63341720.364 ± 2614699.225  ops/s
readwrite     Guava  thrpt   10  43824054.866 ± 1643528.467  ops/s
write_only Caffeine  thrpt   10  43482441.274 ± 2663772.527  ops/s
write_only    Guava  thrpt   10   9052547.148 ±  284201.709  ops/s

# Using volatile updated in background thread [low resolution]
Benchmark  (cacheType) Mode  Cnt      Score         Error      Units
read_only  Caffeine  thrpt   10  168529827.738 ± 10665487.560  ops/s
read_only     Guava  thrpt   10   76641814.064 ±  5860999.940  ops/s
readwrite  Caffeine  thrpt   10  134740502.806 ±  5478623.578  ops/s
readwrite     Guava  thrpt   10   69547375.148 ±  3179356.772  ops/s
write_only Caffeine  thrpt   10   68559250.821 ±  3713187.303  ops/s
write_only    Guava  thrpt   10   10120587.931 ±   200793.621  ops/s
```

johnou:
You mean an impressive explanation, totally worth the read.

johnou:
Think I found all the non-daemon threads which might cause the JVM to hang during benchmarks.

ben-manes:
Thanks! I uploaded a new snapshot manually and pulled your branch, so I'm ready to give that another try.

ben-manes:
There's still a `ScheduledThreadPoolExecutor` somewhere. Unfortunately the thread dumps don't backtrack to where it was created.

ben-manes:
Oh I probably forgot to mvn install again.

johnou:
Smashed it!

Benchmark                               Mode  Cnt        Score       Error  Units
SingleNodeBenchmark.requestThroughput  thrpt  100  3201000.362 ± 46281.930  ops/s

johnou:
Most likely my previous commit wasn't enough, fingers crossed it's ok if you sync once more.

ben-manes:
nice! How does it run without the same-thread executor?

Running fine now. I don't have a baseline to compare though, but will post the numbers when complete

johnou:
The previous number was without the same-thread executor (assuming you are referring to the maintenance thread).

ben-manes:
Yep, I was looking at `ResponseCaching`. What's the reason for leaving that with the executor defined?

johnou:
cloud.orbit.actors.test.CacheResponseTest#testCacheDuration this test here, the CI tests were failing so I suspect it was related to read / write buffering?

ben-manes:
oh most likely. If you can configure the executor for tests only (e.g. via dependency injection) then that avoids it for production. In this case it shouldn't make much of a difference, though.

I don't have a baseline so my numbers probably don't mean much.

```
requestThroughput            thrpt  100  2148790.618 ± 26832.215  ops/s
avgRequestTime_batched        avgt  100        1.048 ±     0.019  us/op
avgRequestTime_singleThread   avgt  100        2.194 ±     0.023  us/op
```

johnou:
You could try it again with Caffeine 2.2.3?

johnou:
Really appreciate the time and effort you have put in to help. If you are ever in Venice area drinks are on me.

ben-manes:
Thanks for your help and patience. Same deal if you're ever in San Francisco.

ben-manes:
Looks like a nice gain. Here's the v2.2.3 numbers.
```
requestThroughput            thrpt  100  2002389.598 ± 23785.454  ops/s
avgRequestTime_batched        avgt  100        1.262 ±     0.024  us/op
avgRequestTime_singleThread   avgt  100        2.404 ±     0.021  us/op
```

If you're happy then I'll cut a release.

johnou:
Very happy, just wondering if the read / write buffer is going to cause any surprises in the other caches.

ben-manes:
Guava used the read buffer, but its an unbounded CLQ with a logical bounding by a counter. That was the easiest way when I was originally figuring out the algorithms (in [CLHM](https://github.com/ben-manes/concurrentlinkedhashmap)). Later I switched to ring buffers, but that never got into Guava. So for the read side this should be better.

CLHM only supports a maximum size, so the write buffer there is used less. Never heard any complaints from Cassandra, et al, so it seemed to be okay.

I think with these fixes there shouldn't be a problem anymore. I'd like to eventually move from an unbounded write buffer to a bounded one to reduce GC (linked nodes) and back-pressure if there is an unexpected runaway like here.

johnou:
Makes sense, thanks again for explaning everything in such detail.

ben-manes:
[Released](https://github.com/ben-manes/caffeine/releases/tag/v2.2.4), but might take a while for Maven Central to sync.

johnou:
Did you manage to get the perform maintenance task to throw an exception? If that's not caught it will stop running, right?

ben-manes:
Pardon? It should keep running, its just not properly logged if an exception occurs. Usually an uncaught exception is handled by the executor by destroying the thread. The oversight of not logging would mask an issue if it occurs, but shouldn't break correctness. Or am I missing something?

ben-manes:
I should rephrase, its not an infinite loop and maybe that's what you thought? The task would stop prematurely, but another one would be scheduled again later. An exception could come from a `CacheWriter#delete()` but that is a rarely used feature so its probable safe enough until the next release.

johnou:
Mixed it up with a TimerTask, if an exception is thrown from its run method it stops running.

ben-manes:
you might be right nonetheless. It might be intelligent and see that the task is complete and can no-op. That might be why I didn't do this optimization before. I better check. Might require another release

ben-manes:
Okay, thanks! I need to make another release, unfortunately.

The executor will ignore the task as complete, so the maintenance work is busted. I'm surprised my tests get by with a single clean-up.

Thankfully there is a `ForkJoinTask#reinitialize()` to reset the internal state prior to submitting. So its an easy fix and won't require regressing to creating garbage.

johnou:
Not just a timer http://docs.oracle.com/javase/6/docs/api/java/util/concurrent/ScheduledExecutorService.html

"If any execution of the task encounters an exception, subsequent executions are suppressed."

ben-manes:
Yes, for any scheduled task that's true. This is scheduled by user operations instead of a timer, so I assumed re-submitting a task (e.g. Runnable) was okay to recycle. An exception wouldn't halt a re-submission. Here, though, the pool is intelligent because tasks are forked and threads may help complete if there is a backlog.

ben-manes:
Committed. Can you take a look and let me know if you see any other oversights? I'll release again if we're good.

johnou:
After doing a bit of reading I wonder if we might need more checks. Java docs state, you have to be sure a task is done and
that no other thread/task is depending on its completion before you can reinitialize/refork.

ben-manes:
The submission to the executor is guarded by a state machine (see `scheduleDrainBuffers`). In addition a tryLock is used to ensure exclusive access as others (primarily `clear()`) will lock to traverse the full data structure. The net result is that the task is submitted only once per completion event, so this should be safe. It isn't forked and there is no one awaiting its completion.

I assumed that JavaDoc was intended to warn that any downstream waiters queue'd actions would be broken. I don't think that applies to here.

johnou:
LGTM! Nice catch.

ben-manes:
Thank you! You caught it :-)

johnou:
I'm not sure if you can mix ForkJoinTask and Runnable like that to get the desired performance boost?

johnou:
At least with my tests when using ForkJoinPool it doesn't even compile (ambiguous method call) execute(ForkJoinTask<?> task) / execute(Runnable task)

ben-manes:
It compiles because I'm using an `Executor` interface, not `ForkJoinPool.commonPool()` directly. Internally it checks to avoid re-wrapping.
```
public void execute(Runnable task) {
    if (task == null)
        throw new NullPointerException();
    ForkJoinTask<?> job;
    if (task instanceof ForkJoinTask<?>) // avoid re-wrap
        job = (ForkJoinTask<?>) task;
    else
        job = new ForkJoinTask.RunnableExecuteAction(task);
    externalPush(job);
}
```

johnou:
You're too quick :) 
breakpoint

ben-manes:
Oh I only realized I could use this trick because I saw the code and the re-wrap comment led me to this optimization =)

ben-manes:
[2.2.5](http://repo1.maven.org/maven2/com/github/ben-manes/caffeine/caffeine/2.2.5) is sync'd to central. Thanks again!

johnou:
@ben-manes cheers!

commons/src/main/java/cloud/orbit/concurrent/Task.java
```diff
@@ -93,6 +93,20 @@ public Task()
...
+        Task.commonPool = commonPool;
+    }
+
+    public static void setSchedulerExecutor(final ScheduledExecutorService schedulerExecutor)
+    {
+        Task.schedulerExecutor = schedulerExecutor;
+    }
+
+    public static void shutdownSchedulerExecutor() {
+        schedulerExecutor.shutdown();
```
johnou: Not 100% sold on this but it prevents the JVM from exiting and if I set a custom thread factory (setting threads to daemon) cloud.orbit.concurrent.test.TaskImplementationTest#testSleep fails to run..

ben-manes:
I think I found another problem with scheduling, so I'm going to release again. I'm not sure if it is benign or not, as it is very weird. Basically the `ForkJoinTask#exec()` needs to return `false` to more reliably re-run in a timely fashion.

ben-manes:
Released. Central has been really slow to sync lately so might take a over an hour. I hope this is the last one for the weekend.

johnou:
@ben-manes they are having some cache invalidation issues.

johnou:
Thanks for informing me about the issue!

ben-manes:
Seems to be the theme of the weekend then :)

[2.2.6](http://repo1.maven.org/maven2/com/github/ben-manes/caffeine/caffeine/2.2.6/) is up on Central.

johnou:
I will investigate a little further soon but it isn't something silly like the fact we never invoke complete when returning true? eg.

```java
    final class PerformCleanupTask extends ForkJoinTask<Void> implements Runnable {
        private static final long serialVersionUID = 1L;
        private final Void NIL = null;

        @Override
        public Void getRawResult() {
            return null;
        }

        @Override
        public void setRawResult(Void v) {}

        @Override
        public boolean exec() {
            try {
                run();
            } catch (Throwable t) {
                logger.log(Level.SEVERE, "Exception thrown when performing the maintenance task", t);
            }
            complete(NIL);
            return true;
        }
```

johnou:
Nope, just invokes the raw result method. 
     public void complete(V value) {
  915           try {
  916               setRawResult(value);
  917           } catch (Throwable rex) {
  918               setExceptionalCompletion(rex);
  919               return;
  920           }
  921           setCompletion(NORMAL);
  922       }

ben-manes:
`doExec` also calls `setCompletion` if done.

```java
    final int doExec() {
        int s; boolean completed;
        if ((s = status) >= 0) {
            try {
                completed = exec();
            } catch (Throwable rex) {
                return setExceptionalCompletion(rex);
            }
            if (completed)
                s = setCompletion(NORMAL);
        }
        return s;
    }
```

It appeared to be a race where that `reinitialize` was sometimes stampeded by FJP. Then the task was viewed as done and the pool ignored the re-submission. The cache's state machine viewed the task as executing so it wouldn't re-submit and draining wasn't occurring. So this seemed to be a race between the task's internal finalization (our code complete) and a new submission.

ben-manes:
It looks like with this change we don't need `reinitialize` anymore because the task isn't complete, so it is resubmitted. This appears to be primarily intended for forking tasks so that the same instance is reused, but in our case it serves for repeated task exection.

johnou:
Worth a code comment above the return false explaining why, wdyt?

ben-manes:
Good idea. There are too many subtle details that are hard to document and test for.

johnou:
@JoeHegarty squashed :+1:

JoeHegarty:
Thanks guys!

PR #151, opened 2016-04-27
Call correct wrapped method.

samskivert:
Previously these were infinitely looping.


samskivert:
FYI: I also signed the CLA. Not sure if you need to wait for EA legal to approve.

JoeHegarty:
Nice catch. CLA looks good from here.

PR #155, opened 2016-04-28
Ensure EA Async is not a transitive dependency in orbit

JoeHegarty:
I want to make sure that Orbit does not pull in async as a transitive dependency. 
Application developers should include async if they want to use it, not be forced to use it.


DanielSperry:
That's looking good.
I think we should merge it.

PR #156, opened 2016-04-29
Removed caffeine Ticker setter from ResponseCaching

johnou:
Motivation:

ResponseCaching exposed a public setter with a reference to a transient dependency.

Modifications:

Replaced custom ticker with fake clock.

Result:

No dependency leakage.

### DIFF
--- a/actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/FakeClock.java
+++ b/actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/FakeClock.java
@@ -66,7 +66,7 @@ public class FakeClock extends Clock
     {
         if (stopped)
         {
-            return millis.incrementAndGet();
+            return millis.get();
         }
         else
         {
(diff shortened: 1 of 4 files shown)


actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/FakeClock.java
```diff
@@ -66,7 +66,7 @@ public long millis()
     {
         if (stopped)
         {
-            return millis.incrementAndGet();
+            return millis.get();
```
johnou: Yes, found a bug in FakeClock! :stuck_out_tongue_winking_eye:

johnou: @ben-manes fyi this was the root cause of the failure for https://github.com/orbit/orbit/pull/144#issuecomment-204718574

PR #160, opened 2016-05-17
Enhance java serialization cloner to support replacing remote objects…

johnou:
… with references.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/cloner/JavaSerializationCloner.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/cloner/JavaSerializationCloner.java
@@ -28,16 +28,19 @@
 
 package cloud.orbit.actors.cloner;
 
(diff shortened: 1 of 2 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/cloner/JavaSerializationCloner.java
```diff
@@ -28,30 +28,33 @@
...
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
 
 /**
  * Java serialization based object cloning implementation.
  */
-public class JavaSerializationCloner implements ExecutionObjectCloner
+public class JavaSerializationCloner extends JavaMessageSerializer implements ExecutionObjectCloner
```
johnou: allows for compatibility with cloud.orbit.actors.runtime.JavaMessageSerializer.ReferenceReplacement.

PR #162, opened 2016-05-18
Kyro cloner to support no-arg classes

johnou:
Motivation:

Kyro cloner does not support no-arg classes.

Modifications:

Configure Kryo to first try to find and use a no-arg constructor and if it fails to do so, fallback to StdInstantiatorStrategy provided by objenesis.

Result:

True immutable objects without no-arg constructors can now be used as Orbit remote objects.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/cloner/KryoCloner.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/cloner/KryoCloner.java
@@ -28,31 +28,40 @@
 
 package cloud.orbit.actors.cloner;
 
+import org.objenesis.strategy.StdInstantiatorStrategy;
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.pool.KryoFactory;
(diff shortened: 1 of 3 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/cloner/KryoCloner.java
```diff
@@ -66,49 +73,21 @@ public Kryo create()
...
-                // Adding support to clone unmodifiable collections as modifiable
-                // collections.
-
-                // It has to be noted that this might not be the expected behavior
-                // in some cases.
-
-                // Without this the unmodifiable collections cause kryo to throw an
-                // exception.. Class cannot be created (missing no-arg constructor).
-
-                kryo.addDefaultSerializer(Collection.class, new CollectionSerializer()
```
johnou: Replaced with the custom serializers below. Instead of a linked x (using double~ the memory!) we will end up with a cloned unmodifiable x :)

johnou:
@JoeHegarty you could try running tests again with the clone message as warn or even error now, I am pretty confident it will copy almost anything you throw at it.

PR #164, opened 2016-06-01
Use HashMap instead of LinkedHashMap for lower overhead.

johnou:


DanielSperry:
I'm cool with this.
No reason comes to mind as to way we would **need** to preserve the order

JoeHegarty:
Sounds good. Will merge. Thanks!

johnou:
@DanielSperry tell you what would be cooler? https://logging.apache.org/log4j/2.x/manual/garbagefree.html

DanielSperry:
hum... you can trade garbage generation for cpu or code complexity.
I think removing linked hashmap is a good thing since it generated more complex (interlinked garbage)

how would you do it? write the headers immediately to a buffer? perhaps in msg pack format?
store them in an object array or list intead of a map?

DanielSperry:
nothing forces us to use the same message class all the time, there could be a message factory. then the application could chose how to implement the message.

dVakulen:
In Orleans message headers will no longer be stored in collection after merge of the  https://github.com/dotnet/orleans/pull/2124.

johnou:
@DanielSperry I was thinking of reusing hashmap with thread local to prevent object creation every request if sticky headers are set but the comment from @dVakulen looks interesting, I'll take a look when I get back to my pc.

johnou:
```c#
if ((headers & Headers.REQUEST_CONTEXT) != Headers.NONE)
{
    var requestData = input.RequestContextData;
    var count = requestData.Count;
    stream.Write(count);
    foreach (var d in requestData)
    {
        stream.Write(d.Key);
        SerializationManager.SerializeInner(d.Value, stream, typeof(object));
    }
}
```

johnou:
@dVakulen thanks for jumping in on this issue, message headers with bitmask would also be ideal for Orbit.

johnou:
Looks like I was too fast posting that comment, Orbit messaging is already using fields for all headers, the headers we are referring to in this issue are custom ones set by the developer Orbit sticky headers == Orleans request context.

johnou:
@dVakulen @DanielSperry what I had meant is changing cloud.orbit.actors.runtime.ActorTaskContext to use a static ThreadContextStack ( http://www.lookatsrc.com/source/org/apache/logging/log4j/spi/DefaultThreadContextMap.java?a=org.apache.logging.log4j:log4j-api ) like log4j2, at the moment for every application request that I process I need to invoke cloud.orbit.actors.runtime.ActorTaskContext#pushNew which creates a new TaskContext object and ConcurrentHashMap (!) just so I can pass in some sticky headers (request context)..

johnou:
Something like this https://github.com/orbit/orbit/compare/master...johnou:feature/thread_request_context?expand=1 .. I would actually get rid of the sticky header properties and dump the entire request context in the invocation pojo, unfortunately the transaction code poses a risk as I am not familiar with it and each context in the stack has its own set of properties so I am not sure if this approach could be a drop in replacement.

johnou:
@DanielSperry I'd love if you could review the above diff and let me know your thoughts, unfortunately I am not familiar with the transactional system but the unit tests passed.. :)

PR #165, opened 2016-06-03
Increase max queue size to 10k, print warning if WaitFreeExecutionSer…

johnou:
…ializer exceeds queue threshold.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/concurrent/WaitFreeExecutionSerializer.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/concurrent/WaitFreeExecutionSerializer.java
@@ -74,7 +74,7 @@ public class WaitFreeExecutionSerializer implements ExecutionSerializer, Executo
         final Task<R> completion = new Task<>();
 
         int queueSize = size.get();
-        if (DEBUG_ENABLED && queueSize >= maxQueueSize / 10)
+        if (DEBUG_ENABLED && queueSize >= maxQueueSize / 10 && queueSize % 1000 == 0)
         {
             logger.debug("Queued " + queueSize + " / " + maxQueueSize + " for " + key);
         }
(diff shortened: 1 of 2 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/concurrent/WaitFreeExecutionSerializer.java
```diff
@@ -74,9 +74,9 @@ public WaitFreeExecutionSerializer(final ExecutorService executorService, Object
         final Task<R> completion = new Task<>();
 
         int queueSize = size.get();
-        if (DEBUG_ENABLED && queueSize >= maxQueueSize / 10)
+        if (WARN_ENABLED && queueSize == maxQueueSize / 10)
```
johnou: to prevent spamming.

JoeHegarty: I'm not sure this is really a warning. Would it make sense to make this info, and then introduce a warning if you exceed the maxSize?

actors/runtime/src/main/java/cloud/orbit/actors/concurrent/WaitFreeExecutionSerializer.java
```diff
@@ -74,7 +74,7 @@ public WaitFreeExecutionSerializer(final ExecutorService executorService, Object
         final Task<R> completion = new Task<>();
 
         int queueSize = size.get();
-        if (DEBUG_ENABLED && queueSize >= maxQueueSize / 10)
+        if (DEBUG_ENABLED && queueSize >= maxQueueSize / 10 && queueSize % 1000 == 0)
```
JoeHegarty: As we discussed, should be maybe go with
```
if (DEBUG_ENABLED && (queueSize >= maxQueueSize / 10 && queueSize % (maxQueueSize / 10) == 0))
```

It would make the logs show the progression of how close we're getting, and scale for any queue size.

johnou:
@JoeHegarty @DanielSperry could we replace the dynamic maxQueueSize / ConcurrentLinkedQueue with a fixed size MpscChunkedArrayQueue from jctools?

PR #168, opened 2016-06-15
Replace Guava and ClassPathSearch with FastClasspathScanner

johnou:
Motivation:

FastClasspathScanner is an uber-fast, ultra-lightweight classpath scanner for Java, Scala and other JVM languages.

Modifications:

Replaced Guava and ClassPathSearch with FastClasspathScanner.

Result:

Less code to maintain.


actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultActorClassFinder.java
```diff
@@ -28,16 +28,73 @@
...
 
 public class DefaultActorClassFinder implements ActorClassFinder
 {
-    private static final ClassPathSearch search = new ClassPathSearch(Actor.class);
+    private static Logger logger = LoggerFactory.getLogger(DefaultActorClassFinder.class);
+
+    private static final Map<Class<?>, Class<?>> concreteImplementations = new ConcurrentHashMap<>();
+
+    static {
+        String basePackage = System.getProperty("orbit.actors.basePackage", "");
```
johnou: might be nice to be able to configure this through stage but I ran out of time.

JoeHegarty:
Thanks @johnou. I'm going to hold off on this one for now as the core issue is fixed, and I'd like some time to review this properly and test it in the scenario where it used to fail.
I agree this is a much better implementation, just want to verify everything first.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultActorClassFinder.java
```diff
@@ -28,16 +28,73 @@
...
+        String[] scanSpec = new String[]{};
+        if (!basePackage.isEmpty())
+        {
+            String[] basePackages = basePackage.split(",");
+            Set<String> tmp = new HashSet<>(basePackages.length);
```
johnou: LHS

actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultActorClassFinder.java
```diff
@@ -28,16 +28,73 @@
...
+            for (String s : basePackages)
+            {
+                if (!s.trim().isEmpty())
+                {
+                    tmp.add(s);
+                }
+            }
+            tmp.add("cloud.orbit"); // internal actors
+            scanSpec = tmp.toArray(new String[tmp.size()]);
```
johnou: Verify -com.sun

johnou: For efficiency, system, bootstrap and extension jarfiles like rt.jar (i.e. the jarfiles distributed with the JRE) are always blacklisted, as are package prefixes java and sun, i.e. they are never scanned. If you put custom classes into the lib/ext directory in your JRE folder (which is a valid but rare way of adding jarfiles to the classpath), they will be ignored by association with the JRE.

PR #174, opened 2016-06-21
Performance regression fixes

johnou:


johnou:
@lukehutch seeing pretty bad performance with fast-classpath-scanner in our integration environment when everything is packaged into jars etc. scanning is taking up to a minute, when run from the IDE it's maybe a few seconds.. maybe I am using the API incorrectly or potentially some recursive scanning bug?

lukehutch:
Interesting, scanning should never take that long in any reasonable classpath. Please call .verbose() before scanning, and compare the output. There is scan time info in the log. Also you can compare the classpath output in the log by the different scans. Also please let me know if the performance regression happened between 1.9.18 and 1.9.19, there were some major internal changes in 1.9.19.

lukehutch:
N.B. the change between 1.9.18 and 1.9.19 could cause slowdown in cases where you have a lot of implemented interfaces/superinterfaces and a lot of subclasses of a given class. If you find that downgrading to 1.9.18 fixes the performance regression, I'll optimize that part of the code.

johnou:
1.9.20 FastClasspathScanner: *** Time taken by .scan(): 78806 ms ***
1.9.18 FastClasspathScanner: *** Time taken by .scan(): 121116 ms ***

lukehutch:
@johnou Strange, do you have those numbers backwards? I thought you were saying that 1.9.20 was slower than older versions? But the numbers show it's 35% faster.

I just released 1.9.21, which should be significantly faster than 1.9.20 in cases where there are classes with many subclasses, some or all of which implement many interfaces.

johnou:
One curious thing I have noticed is that somewhere it must be resolving a symlink but then double scanning jars..

FastClasspathScanner: Scanned jarfile /home/xxx/builds/develop-SNAPSHOT/libs/spring-tx-4.2.6.RELEASE.jar in 1369 msec
FastClasspathScanner: Scanned jarfile /home/xxx/builds/latest/libs/spring-tx-4.2.6.RELEASE.jar in 2 msec

/home/xxx/builds/latest -> /home/xxx/builds/develop-SNAPSHOT

lukehutch:
@johnou if classes with the same name are listed twice on the classpath, the second and subsequent instance of that class name is shadowed by the first (meaning the class is not scanned, in keeping with the Java classpath spec). This is why the second jar in your example is scanned in 2 msec -- all the classfiles in the zipfile directory listing were already scanned the first time the jar was listed in the classpath. (This is why it took 1369 msec the first time, the actual classfile contents were read, not just the zipfile header.)

johnou:
Eugh, my apologies, bad wording on my part. Regression with Orbit bootstrap after swapping over to fast class path scanner.

lukehutch:
@johnou ?? I don't understand your last comment. See also my question two comments before this one.

johnou:
1. My first message to you was worded badly, sounded like there was a regression in fastclasspath library but what I meant is bootstrap time for Orbit.
2. No, I am pretty sure 1.9.20 scanned the classpath faster but I can compare that again.
3. 1.9.21 FastClasspathScanner: *** Time taken by .scan(): 70711 ms ***
4. Looking into this now

johnou:
@lukehutch comparing IDE and integration verbose output I can see that in integration it is logging lines like the following which are not present in my IDE..

FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSMWW has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.WriteOrderDeque
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSMWW has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.WriteOrderDeque
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSMWW has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.Node
FastClasspathScanner: Found class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSMWWR with superclass cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSMWW
FastClasspathScanner: Found class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSR with superclass cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiS
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSR has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.Ticker
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSR has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.WriteBuffer
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSR has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.WriteBuffer
FastClasspathScanner: Found class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSW with superclass cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiS
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSW has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.Ticker
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSW has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.WriteOrderDeque
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSW has a field with type or type parameter cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.WriteOrderDeque
FastClasspathScanner: Class cloud.orbit.runtime.shaded.com.github.benmanes.caffeine.cache.LocalCacheFactory$SILiSW has a field with type or type parameter

lukehutch:
I see, thanks for explaining, and for testing 1.9.21.

Is there anything you can change about the scanning spec (whitelist / blacklist specification) to speed up scanning in Orbit? You're probably scanning a lot of stuff you don't need to.

It still concerns me though that the scanning time is so different between your IDE and your integration environment. How is the classpath different between those two scenarios?

It also concerns me that the total scanning time is 70 seconds. That's a lot of time even if you're trying to read a few hundred jarfiles. Your classpath must be enormous! What's the approximate number of jarfiles on your classpath, and the total compressed size of the jars?

The log entries you pasted are from 1.9.21, not from git master. If you use the version in git master, and produce verbose logs from the IDE and the integration environment, you should be able to more easily spot the difference between the two environments.

johnou:
Yes configurable scanning spec is possible but out of the box configuration is entire classpath and I am pretty sure it was not taking this long with Guava, I will build a snapshot artifact of fastclasspath now.

lukehutch:
I'd be very interested to see the runtimes for Guava, if you can run that again. Do you know if Guava is searching just a subset of the classpath? Is there a way to generate debug output there, to compare? I don't have experience with Guava.

It's possible Guava is using a faster unzip library than the default Java one -- that is known to be very slow, and it is not thread-safe. I have considered replacing the library, but haven't done it yet. I'll create a bug for that, and will take a look at what Guava uses.

What specific methods are you using in FastClasspathScanner? Are you looking just for annotations, or something else?

johnou:
Deploying snapshot artifact to integration environment now, use case is that we need to find all interfaces extending cloud.orbit.actors.Actor and then find all concrete classes implementing those interfaces. Current source I am running for that can be found here in the class constructor https://github.com/johnou/orbit/blob/0364ff25d147588d5cc5495135797af03a9e615c/actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultActorClassFinder.java I will try and find some time to reimplement the ActorClassFinder with guava and compare.

johnou:
From IDE:
2016-06-23T10:33Z	FastClasspathScanner	- Built class graph in 0.129220 sec
2016-06-23T10:33Z	FastClasspathScanner	Finished .scan() in 7.662825 sec
2016-06-23T10:33Z	FastClasspathScanner	Parsed 28776 classfiles on classpath

johnou:
Integration environment:
2016-06-23T10:35Z       FastClasspathScanner    - Built class graph in 5.819232 sec
2016-06-23T10:35Z       FastClasspathScanner    Finished .scan() in 87.235730 sec
2016-06-23T10:35Z       FastClasspathScanner    Parsed 34628 classfiles on classpath

lukehutch:
Wow, big difference. Can you please include a pastebin of the logs from these two environments?

johnou:
I need to head out for lunch now but is there an email I can send the contents? May be some information I don't want everyone to see :)

lukehutch:
Yes, my gmail address is luke.hutch. I'm about to send a comment with a probable fix though (couple of mins)

lukehutch:
OK, I see a couple of problems with the implementation that will be causing a lot of the problems:

(1): MatchProcessors are all run in the same pass. So you can't add one MatchProcessor (here, the one that populates clazzInterfaces), and then depend on its result in a second MatchProcessor within the same scan operation (here, the one with the loop "for (Class<?> clazzInterface : clazzInterfaces)").

Consider what this is doing: every time the first MatchProcessor adds an Actor to clazzInterfaces, the second MatchProcessor has one extra item to iterate through when it hits an Actor. So you actually created an O(N^2) algorithm. Additionally, it will fail to find some subclasses, because clazzInterfaces won't be fully populated before the first time the second MatchProcessor is triggered.

The quick solution to this would be to create two FastClasspathScanner instances, and run the first scan with the first MatchProcessor, then when it has finished, run the second scan with the second MatchProcessor. **However**, there is a much easier way!:

(2): You don't need to find all subinterfaces of an interface before searching for each one of those. FastClasspathScanner already does that for you. Just look for classes that implement cloud.orbit.actors.Actor. If a class implements a subinterface of Actor, then it still implements Actor. Here you can see the code that finds all subinterfaces of the given interface you're trying to match, then takes a union of all the subclasses of each class that implements the interface: https://github.com/lukehutch/fast-classpath-scanner/blob/master/src/main/java/io/github/lukehutch/fastclasspathscanner/classgraph/ClassGraphBuilder.java#L243 

So all you need to do is use .matchClassesImplementing(Actor.class, c -> /*...*/).

One more thing though: I'm still concerned about the difference in "Built class graph in" times in the two instances -- they should be approximately the same, given that the number of classfiles parsed is approximately the same. So I'd still be curious to see those logs. Thanks!

lukehutch:
So based on a quick cut of the unnecessary code, it looks like here's all you need to do:

```
        FastClasspathScanner scanner = new FastClasspathScanner(scanSpec).matchClassesImplementing(Actor.class, clazzImplementation -> {
            if (clazzImplementation.getSimpleName().toLowerCase(Locale.ENGLISH).startsWith("abstract"))
            {
                return;
            }
            Class<?>[] implementationInterfaces = clazzImplementation.getInterfaces(); // check interfaces directly to support inheritance
            if (implementationInterfaces.length == 0)
            {
                return;
            }
            Class<?> old = concreteImplementations.put(clazzInterface, clazzImplementation);
            if (old != null)
            {
                logger.warn("Multiple actor implementations found for " + clazzInterface);
            }
        });
        scanner.scan();
```

Using the above code, please show me the scanning times again for the git snapshot you built, and for 1.9.18. And please send these two logs via pastebin (not the logs generated from the original O(N^2) code). I'll use this to verify that there's not something funky going on that is adding a slowdown in the classgraph building stage. Thank you!

johnou:
So after lunch I took your advice and revisited the class finder..

```java
    public DefaultActorClassFinder(final String... actorBasePackages)
    {
        String[] scanSpec = extractScanSpec(actorBasePackages);

        List<Class<?>> clazzImplementations = new ArrayList<>();
        long start = System.currentTimeMillis();
        FastClasspathScanner scanner = new FastClasspathScanner(scanSpec).matchClassesImplementing(Actor.class, candidate -> {
            if (candidate.getSimpleName().toLowerCase(Locale.ENGLISH).startsWith("abstract"))
            {
                return;
            }
            clazzImplementations.add(candidate);
        });
        scanner.verbose().scan();
        for (Class<?> clazzImplementation : clazzImplementations)
        {
            Class<?>[] implementationInterfaces = clazzImplementation.getInterfaces(); // check interfaces directly to support inheritance
            if (implementationInterfaces.length == 0)
            {
                return;
            }
            for (Class<?> implementationInterface : implementationInterfaces)
            {
                if (implementationInterface == Remindable.class)
                {
                    continue;
                }
                if (Actor.class.isAssignableFrom(implementationInterface))
                {
                    Class<?> old = concreteImplementations.put(implementationInterface, clazzImplementation);
                    if (old != null)
                    {
                        logger.warn("Multiple actor implementations found for " + implementationInterface);
                    }
                }
            }
        }
        if (logger.isDebugEnabled())
        {
            logger.debug("Took " + (System.currentTimeMillis() - start) + "ms to scan for Actor implementations.");
        }
    }
```

johnou:
IDE:
2016-06-23T12:36Z	FastClasspathScanner	- Built class graph in 0.128453 sec
2016-06-23T12:36Z	FastClasspathScanner	Finished .scan() in 7.870147 sec
2016-06-23T12:36Z	FastClasspathScanner	Parsed 28776 classfiles on classpath

integration:
2016-06-23T12:45Z       FastClasspathScanner    - Built class graph in 4.364496 sec
2016-06-23T12:45Z       FastClasspathScanner    Finished .scan() in 102.376371 sec
2016-06-23T12:45Z       FastClasspathScanner    Parsed 34628 classfiles on classpath

johnou:
By removing verbose time dropped significantly as expected.

2016-06-23 13:15:26,189 DEBUG [main] (cloud.orbit.actors.runtime.DefaultActorClassFinder) Took 51505ms to scan for Actor implementations.

johnou:
Spring actor class finder https://gist.github.com/johnou/03c3677cd72f40d3778ea1f7b39df9d3

IDE
2016-06-23 16:00:20,052 INFO  [main] (com.sulake.h4k.common.orbit.SpringActorClassFinder) Took 13437ms to scan for Actor implementations. 

integration environment
8+ minutes..

"main" #1 prio=5 os_prio=0 tid=0x00007fdd3000d000 nid=0x5958 runnable [0x00007fdd39094000]
   java.lang.Thread.State: RUNNABLE
        at java.lang.String.intern(Native Method)
        at java.util.jar.Attributes$Name.<init>(Attributes.java:466)
        at java.util.jar.Attributes.putValue(Attributes.java:168)
        at java.util.jar.Attributes.read(Attributes.java:421)
        at java.util.jar.Manifest.read(Manifest.java:251)
        at sun.security.util.SignatureFileVerifier.processImpl(SignatureFileVerifier.java:251)
        at sun.security.util.SignatureFileVerifier.process(SignatureFileVerifier.java:238)
        at java.util.jar.JarVerifier.processEntry(JarVerifier.java:316)
        at java.util.jar.JarVerifier.update(JarVerifier.java:228)
        at java.util.jar.JarFile.initializeVerifier(JarFile.java:383)
        at java.util.jar.JarFile.getInputStream(JarFile.java:450)
        - locked <0x00000000fb60c4c8> (a sun.net.www.protocol.jar.URLJarFile)
        at sun.net.www.protocol.jar.JarURLConnection.getInputStream(JarURLConnection.java:162)
        at org.springframework.core.io.UrlResource.getInputStream(UrlResource.java:168)
        at org.springframework.core.type.classreading.SimpleMetadataReader.<init>(SimpleMetadataReader.java:50)
        at org.springframework.core.type.classreading.SimpleMetadataReaderFactory.getMetadataReader(SimpleMetadataReaderFactory.java:98)
        at org.springframework.core.type.classreading.CachingMetadataReaderFactory.getMetadataReader(CachingMetadataReaderFactory.java:102)
        - locked <0x00000000fdcf5470> (a org.springframework.core.type.classreading.CachingMetadataReaderFactory$1)
        at org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider.findCandidateComponents(ClassPathScanningCandidateComponentProvider.java:279)

lukehutch:
Just to verify, were the log files you emailed me generated after changing your code?

It's still taking much longer than expected to run this scan, which is really surprising. Will take a look at the logs later today.

What is the backtrace-looking thing you pasted in your previous comment? Is that some sort of profiling info?

Do you happen to have any remote jars on the classpath?

johnou:
Stacktrace is spring stuck trying to chew through the classpath with no "scanspec" set. I still need to benchmark the original guava code on the integration environment. I also plan to try and replicate this on my rig, the integration server seems a little overloaded so I want to rule that out.

johnou:
Nope, no remote jars or network mounts etc.

johnou:
I'm pretty confident the killer was the original O(N^2) code with no scanspec but I would like to ensure it smashes the guava implementation like it did with spring.

lukehutch:
I plotted a histogram of the times -- 10% of the individual classfile parsing steps have a slowdown of 10-1000x on the integration server compared to the IDE, so the slowdown is very spiky. It may just be the integration server choking (e.g. swapping to disk constantly, or being under heavy load). Would be very interested to see the comparison of integration environment vs. IDE on the same machine.

johnou:
@lukehutch after looking into the original implementation (including guava) I am banging my head on the table, it doesn't parse the entire classpath like I am doing with fastclasspathscanner, instead it takes all classnames and processes them lazily by guessing the implementation class based on the name of the class.. if you are ever in the Venice area let me know and I'll be happy to buy you a few cold beers.

johnou:
@JoeHegarty let me know if you are happy with this or you would like me to also create a LazyActorClassFinder variant.

lukehutch:
@johnou That's quite surprising (is it Guava that does that, or the code that calls Guava?), but the lack of a good way to do classpath scanning properly is exactly why I ended up creating FastClasspathScanner!

I'd still like to know why the scanning times are so spiky on the integration server (with slowdowns of 10-1000x on 10% of the classes). For example, I wonder if it's always the same classes that are slow. Can you please email me another timed log, so I can check this? If it's different classes every time, then it's probably the server being swamped.

Alternatively, is there a way for you to duplicate the integration environment exactly, but run the scan on a completely unloaded machine?

The thing I would like to eliminate is some sort of craziness in the class graph that is causing massive slowdowns. The IDE scans 28776 classes, but the integration environment scans 34628 classes, so the classpaths are not exactly the same. Building the class graph took 340x longer in the integration environment, for the times you sent me. That particular step operates directly in memory, and it is just finding the upstream and downstream transitive closures from each class, so that sort of slowdown is ridiculous if the hardware is approximately the same between the two environments, and if the slowdown is not due to machine loading.

lukehutch:
I just optimized classpath scanning for cases like yours, with large classpaths and small numbers of matchers or queries. (The class graph construction step now happens on-demand, and only for requested classes, which should save you a few seconds at least.) I haven't pushed out a new release yet.

I'm also looking at ways to speed up unzipping of jarfiles -- I will probably switch to NIO filesystem support, rather than java.util.zip. This will require a bit of tricky code surgery.

actors/runtime/src/test/java/cloud/orbit/actors/extensions/test/ConstructionTest.java
```diff
@@ -54,13 +54,13 @@ public void testDefaultConstruction() throws Exception {
...
 
-            stage.start();
+            stage.start().join();
 
             final TestConstruction actor = Actor.getReference(TestConstruction.class, "0");
             assertEquals(DEFAULT_ID, actor.getId().get());
 
         } finally {
-            stage.stop();
+            stage.stop().join();
```
johnou: @jpennell tut tut.

johnou:
@lukehutch I will try and get you the new logs with latest snapshot later tonight along with setting up my machine like the integration environment.

JoeHegarty:
Thanks for all the efforts @johnou and @lukehutch. 
I've merged the change as it is a vast improvement over what we had in 0.9.4.

I know you still plan to try out the latest snapshot etc, please feel free to just open a new PR, any further improvement would be greatly appreciated. Just wanted to get out the great improvements you've already made to our users.

lukehutch:
The main remaining optimization I can implement is by switching to something that can unzip faster than java.util.zip (I'm hoping NIO will do this). But don't keep get too hopeful ;)  I think FastClasspathScanner is now running within a small constant factor of the fastest theoretical limit. 

It's conceivable there's something else funky going on though, so I'll check the next set of logs. The evidence points to a heavily-loaded integration server.

The main way you can speed up scanning is by proper use of whitelists and blacklists.

johnou:
Must be hitting some IO limit on the integration environment..

2016-06-25T00:12Z	FastClasspathScanner	Finished .scan() in 32.100488 sec
2016-06-25T00:12Z	FastClasspathScanner	Parsed 34705 classfiles on classpath

vs my rig
2016-06-25T00:26Z       FastClasspathScanner    Finished .scan() in 9.521939 sec
2016-06-25T00:26Z       FastClasspathScanner    Parsed 34705 classfiles on classpath

johnou:
specifying package on my rig
2016-06-25T00:36Z       FastClasspathScanner    Finished .scan() in 1.212041 sec
2016-06-25T00:36Z       FastClasspathScanner    Parsed 2950 classfiles on classpath

johnou:
>> That's quite surprising (is it Guava that does that, or the code that calls Guava?)
code that calls Guava.. https://github.com/orbit/orbit/pull/175/files

lukehutch:
OK, so only considering your rig, is it true that the ratio in time taken to scan on the IDE to the time taken in the integration environment is roughly the same as the ratio in the number of classes scanned in each environment?  i.e. is the average time taken to scan one class roughly the same in both environments on your rig?

johnou:
From IDE to cli on the same machine there was only a few seconds difference but it has an SSD drive so cannot be compared directly to the integration environment. I'd still like to try it on Linux with symlinks and a "normal" drive to really replicate the integration environment.

johnou:
@lukehutch apologies for not getting back to you sooner. I managed to setup a new environment identical to the integration, there is definitely something fishy with classpaths which include symlinks..

direct with verbose
2016-06-29T09:09Z       FastClasspathScanner    Finished .scan() in 55.743153 sec
2016-06-29T09:09Z       FastClasspathScanner    Parsed 34736 classfiles on classpath

symlink with verbose
2016-06-29T09:11Z       FastClasspathScanner    Finished .scan() in 97.918884 sec
2016-06-29T09:11Z       FastClasspathScanner    Parsed 34736 classfiles on classpath

[johno@dev-1 fs]$ java -cp "build/libs/*:build/server/test-service.jar" OrbitStarter
2016-06-29 12:12:19,610 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 4524ms to scan for Actor implementations.
[johno@dev-1 fs]$ java -cp "build/libs/*:build/server/test-service.jar" OrbitStarter
2016-06-29 12:12:27,188 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 4414ms to scan for Actor implementations.
[johno@dev-1 fs]$ java -cp "build/libs/*:build/server/test-service.jar" OrbitStarter
2016-06-29 12:12:33,058 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 4326ms to scan for Actor implementations.
[johno@dev-1 fs]$ java -cp "latest/libs/*:latest/server/test-service.jar" OrbitStarter
2016-06-29 12:12:50,408 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 6983ms to scan for Actor implementations.
[johno@dev-1 fs]$ java -cp "latest/libs/*:latest/server/test-service.jar" OrbitStarter
2016-06-29 12:12:58,971 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 6832ms to scan for Actor implementations.
[johno@dev-1 fs]$ java -cp "latest/libs/*:latest/server/test-service.jar" OrbitStarter
2016-06-29 12:13:07,376 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 6897ms to scan for Actor implementations.
[johno@dev-1 fs]$ java -cp "build/libs/*:build/server/test-service.jar" OrbitStarter
2016-06-29 12:13:15,998 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 4210ms to scan for Actor implementations.
[johno@dev-1 fs]$ ls -lah
total 52M
drwxr-xr-x   3 johno johno 4.0K Jun 29 12:11 .
drwxr-xr-x 323 johno root    20K Jun 29 12:11 ..
drwxr-xr-x   4 johno johno 4.0K Jun 29 10:01 build
lrwxrwxrwx   1 johno johno    5 Jun 29 10:01 latest -> build

johnou:
updated timetaken when piping output directly into file when in verbose mode.

johnou:
[johno@dev-1 fs]$ wc -l symlink-verbose.log
329568 symlink-verbose.log
[johno@dev-1 fs]$ wc -l direct-verbose.log
190003 direct-verbose.log

johnou:
Classpath elements contains both symlinks and the resolved path, must be scanning every artifact twice.

lukehutch:
OK, I just pushed a change to FastClasspathScanner that resolves symlinks, and then ensures that files, directories and jars are only read once. This should cause the time for ide vs. integration to match on your machine. Can you please verify this? (I haven't pushed out a new release yet, so you'll need to build FastClasspathScanner from source.)

(The reason I wasn't originally resolving symlinks is that the directory structure has to match the package structure in Java, and this can break after resolving symlinks. Now I resolve them once, just to check for duplicates, which should fix your usecase.)

The next release will hopefully scan zipfiles faster, assuming that the NIO unzip code is faster than java.util.zip (I haven't finished this change yet though, so I can't test if it's faster).

johnou:
wrt NIO unzip code, might be a good idea to keep an eye on memory too?

johnou:
@lukehutch that's it! Latest snapshot of FastClasspathScanner no longer takes double~ the scan time when classpath uses symlinks.

lukehutch:
Good call. You have a much larger classpath than I will likely ever work
with though, so it would be great if you could please test this once I make
the NIO commit.

I made changes a few commits ago that dramatically decrease the memory
requirements of FastClasspathScanner, but the Java unzip methods have their
own memory requirements (I assume these are not more than the zip directory
tree structure plus some buffers though).

There will be two separate NIO commits: one to use NIO FileSystems (and a
TreeWalker), which will greatly simplify the RecursiveScanner code. Then
one to multithread the NIO scanning (which will increase memory overhead
but decrease scan time.) Should get these done in the next 2-3 weeks or so.

lukehutch:
@johnou Excellent! Thanks for reporting this, and thanks for all the testing, sending me logs etc.

lukehutch:
I'll push out a new release with these changes and the single-threaded NIO unzip rework in the next couple of weeks, then the multithreaded scanning support will go into version 2.0, since that will require MatchProcessors to add "synchronized" if they are not threadsafe (which is a slightly dangerous change to make -- or maybe I'll require a separate ".parallel()" call like the Java 8 Streams API to explicitly enable parallel scanning).

johnou:
Absolutely, just ping me when you make the NIO commit. Really appreciate the effort you are putting into the library.

lukehutch:
I just published FastClasspathScanner 1.10.0, which switches to (single-threaded) NIO scanning of jarfiles. In my own testing (with a small classpath), it seems a bit faster than the java.util.zip scanner, but it would be great if you could please let me know if it's faster for you. Please also let me know if you see any regressions, since this was a pretty major change.

lukehutch:
Sorry, make that 1.10.2 :-)

johnou:
To my surprise single-threaded NIO seems to be performing worse than the previous releases.

prenio without package
2016-07-04 10:42:18,620 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 3040ms to scan for Actor implementations. 
2016-07-04 10:42:20,376 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1754ms to scan for Actor implementations. 
2016-07-04 10:42:22,144 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1768ms to scan for Actor implementations. 
2016-07-04 10:42:23,793 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1649ms to scan for Actor implementations. 
2016-07-04 10:42:25,488 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1695ms to scan for Actor implementations. 
2016-07-04 10:42:27,138 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1650ms to scan for Actor implementations. 
2016-07-04 10:42:28,714 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1576ms to scan for Actor implementations. 
2016-07-04 10:42:30,685 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1971ms to scan for Actor implementations. 
2016-07-04 10:42:32,271 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1585ms to scan for Actor implementations. 
2016-07-04 10:42:33,837 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1565ms to scan for Actor implementations. 
2016-07-04 10:42:35,561 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1724ms to scan for Actor implementations. 
2016-07-04 10:42:37,097 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1536ms to scan for Actor implementations. 
2016-07-04 10:42:38,732 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1635ms to scan for Actor implementations. 
2016-07-04 10:42:40,460 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1728ms to scan for Actor implementations. 
2016-07-04 10:42:41,916 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1456ms to scan for Actor implementations. 
2016-07-04 10:42:43,411 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1495ms to scan for Actor implementations. 
2016-07-04 10:42:45,154 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1743ms to scan for Actor implementations. 
2016-07-04 10:42:46,593 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1439ms to scan for Actor implementations. 
2016-07-04 10:42:48,230 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1637ms to scan for Actor implementations. 
2016-07-04 10:42:49,712 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 1482ms to scan for Actor implementations. 

nio without package
2016-07-04 10:39:23,973 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 5027ms to scan for Actor implementations. 
2016-07-04 10:39:26,993 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 3018ms to scan for Actor implementations. 
2016-07-04 10:39:29,873 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2880ms to scan for Actor implementations. 
2016-07-04 10:39:32,666 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2793ms to scan for Actor implementations. 
2016-07-04 10:39:35,361 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2695ms to scan for Actor implementations. 
2016-07-04 10:39:38,073 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2712ms to scan for Actor implementations. 
2016-07-04 10:39:40,913 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2840ms to scan for Actor implementations. 
2016-07-04 10:39:43,687 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2774ms to scan for Actor implementations. 
2016-07-04 10:39:46,446 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2758ms to scan for Actor implementations. 
2016-07-04 10:39:49,301 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2855ms to scan for Actor implementations. 
2016-07-04 10:39:52,189 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2888ms to scan for Actor implementations. 
2016-07-04 10:39:55,034 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2845ms to scan for Actor implementations. 
2016-07-04 10:39:57,857 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2823ms to scan for Actor implementations. 
2016-07-04 10:40:00,710 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2853ms to scan for Actor implementations. 
2016-07-04 10:40:03,575 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2865ms to scan for Actor implementations. 
2016-07-04 10:40:06,447 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2872ms to scan for Actor implementations. 
2016-07-04 10:40:09,360 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2913ms to scan for Actor implementations. 
2016-07-04 10:40:12,010 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2650ms to scan for Actor implementations. 
2016-07-04 10:40:14,720 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2710ms to scan for Actor implementations. 
2016-07-04 10:40:17,455 DEBUG [ForkJoinPool.commonPool-worker-1] (cloud.orbit.actors.runtime.FastActorClassFinder) Took 2735ms to scan for Actor implementations.

lukehutch:
I spent many hours trying to figure out why the NIO version was slower, and as far as I can tell, the NIO code triggers some sort of slow path in the Linux kernel when NativeThread.current() calls pthread_current(). This can add 40% overhead per file unzipped. I filed an OpenJDK bug on this.

While debugging this though, I realized that the NIO code for ZipFileSystem actually uses java.util.zip under the hood, so there is actually no benefit to using NIO. Additionally, I found that some of the newer Java 7 APIs, such as Path.toUri(), and some of the static Files methods, are horribly inefficient when used with ZipFileSystem. Consequently, I reverted many of my recent changes, and then optimized what remained using a profiler. Scanning should be fast again now, as of 1.10.4 (just released).

Please test 1.10.4, and let me know how it works for you. It should be no slower than 1.9.21, and may be a little faster on average.

(The only additional thing that can be done is to implement parallel scanning. This will be complicated, but I think I have figured out a way to do this, so hopefully that will be in a future release.)

lukehutch:
@johnou FastClasspathScanner now supports parallel scanning (in the latest version in git -- I haven't pushed out a new release yet). This new version is probably operating at close to the theoretical limit for how fast classpath scanning can go, given that we're overlapping reading from storage with decompression (i.e. I believe we're limited by disk read speeds beyond this point). I would appreciate your testing of this code -- you should see some good speedups.

In my own testing, the parallel version in git is 30-60% faster than the older single-threaded version (1.9.21) for a moderately large classpath (generally it's 30% faster for the first scan, and 60% faster for subsequent scans once the files are all cached). This is achieved using 5 threads, which gave the best speedup on my machine (my machine should be fairly typical of a modern laptop or desktop). Increasing the number of threads causes slowdown due to filing system resource contention. The threads are all about 50% busy decompressing data, and spend 50% of their time waiting for the next chunk to arrive to be decompressed.

(You can tweak the number of threads if you want by changing `io.github.lukehutch.fastclasspathscanner.scanner.RecursiveScanner.NUM_PARALLEL_TASKS`, though I don't expect you'll get dramatically better performance for values other than the default of 5.)

lukehutch:
@johnou I have spent a lot of time profiling FastClasspathScanner, and fixed all the profiling hotspots I could find. Performance in 1.90.4 should work well for you. Please let me know how it works in your environment.

johnou:
@lukehutch smashed it! I think the library really lives up to its name now. Did you try different executors? At the moment number of threads can be configured but maybe custom executor service would be a nice touch, I would be interested in knowing how fork join pool compares, in fact fork join pool common pool might be a suitable default? Assuming it's not slow compared to fixed thread pool of cpu count plus 2..

lukehutch:
Thanks for testing, what times (or speedups) are you seeing now?

I do see a large overhead in the profiler from using the executor service, but I don't know that switching to a different executor will help -- hard to say if the overhead is thread overhead or service overhead. The threading is simple enough though that maybe I'll just switch to using bare threads.

johnou:
I will run more tests this afternoon, travelling at the moment and development from a tablet sucks.

lukehutch:
I found a way to replace the ExecutorService with a custom multithreading approach that increases scanning speed by a further 8% or so. It is in 1.90.6.

PR #177, opened 2016-07-01
Remove dead code from prototype.

johnou:


johnou:
@DanielSperry PTAL.

JoeHegarty:
I'm not sure I understand the implication of this, so I'd like @DanielSperry to comment ideally. 
Does it appear to work @johnou? Have you tried it in an actual deployed distributed build?

actors/infinispan-cluster/src/main/java/cloud/orbit/actors/cluster/JGroupsClusterPeer.java
```diff
@@ -184,15 +183,11 @@ public void receive(final Message msg)
...
-
                     cacheManager = new DefaultCacheManager(globalConfigurationBuilder.build(), builder.build());
 
-                    ConfigurationBuilder builder2 = new ConfigurationBuilder();
-                    builder2.clustering().cacheMode(CacheMode.REPL_SYNC);
-                    cacheManager.defineConfiguration("clusterTopologyCache", builder2.build());
-
                     // need to get a cache, any cache to force the initialization
-                    cacheManager.getCache("clusterTopologyCache");
+                    cacheManager.getCache("distributedDirectory");
```
DanielSperry: I think this was the only essential part.
You need to touch infinispan to get it to start working.

the other cache was never used, it could be but
ClusterPeer listens to jgroups view events and updates a list of active nodes.

Now... my initial goal was to put the node capabilities, 
with a list of all the actors the node supports, in infinispan. 
Never did because we kept locating the actors in the class path instead of making the actor list a configuration parameter.

DanielSperry:
It looks good, but as @JoeHegarty pointed out it would be safer to run a real cluster with the change. before merging it to master.
A go through the clusterpeer tests to see if they exercise this correctly would also be nice.

johnou:
I will run this Monday on our integration environment.

johnou:
Built a snapshot using this branch, integration tests passed.

PR #178, opened 2016-07-14
Dirty local caches.

johnou:
Dirty local cache
Motivation:

There are some cases where nodes will not be informed if an actor has moved or deactivated.

Modifications:

Notify invocation node if actor has moved.
Clear actor reference from local cache when actor deactivates.
Unit test for checking dirty local cache.

Result:

Correct behaviour when using OnlyIfActivated annotation in multi-JVM cluster.

### DIFF
--- a/actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
+++ b/actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
@@ -28,15 +28,15 @@
 
 package cloud.orbit.actors.test;
 
-
-import cloud.orbit.actors.Actor;
-import cloud.orbit.actors.Stage;
-import cloud.orbit.actors.test.actors.SomeActor;
-import cloud.orbit.actors.test.actors.StatelessThing;
-
 import org.junit.Assert;
 import org.junit.Ignore;
 import org.junit.Test;
+
+import cloud.orbit.actors.Actor;
+import cloud.orbit.actors.Stage;
+import cloud.orbit.actors.runtime.RemoteReference;
+import cloud.orbit.actors.test.actors.SomeActor;
+import cloud.orbit.actors.test.actors.StatelessThing;
 
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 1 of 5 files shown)


johnou:
This fix took more work than expected, development from a tablet is not enjoyable.

actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
```diff
@@ -113,6 +113,25 @@ public void cleanupTest() throws ExecutionException, InterruptedException
...
+        Stage stage2 = createStage();
+        SomeActor actor1 = Actor.getReference(SomeActor.class, "1");
+
+        stage1.bind();
+        actor1.sayHello("huuhaa").join();
+        stage2.bind();
+        assertNotNull(actor1.sayFarewell().join());
+        clock.incrementTimeMillis(TimeUnit.HOURS.toMillis(1));
+        stage1.cleanup().join();
+        Thread.sleep(1500); // give stage2 time to process notification
```
johnou: Remove and moved are one way messages so there is no reliable way to know when the message has been processed, we cannot poll the actor either as that would simply trigger activation.

actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
```diff
@@ -113,6 +121,80 @@ public void cleanupTest() throws ExecutionException, InterruptedException
...
+        assertNotNull(actor1.sayHelloOnlyIfActivated().join());
+        clock.incrementTimeMillis(TimeUnit.HOURS.toMillis(1));
+        stage1.cleanup().join();
+        eventuallyTrue(2000, () -> stage1.locateActor(RemoteReference.from(actor1), false).get() == null);
+        assertNull(actor1.sayHelloOnlyIfActivated().join());
+        dumpMessages();
+    }
+
+    @Test
+    public void notifyOnActorMove() throws ExecutionException, InterruptedException, TimeoutException
```
johnou: @DanielSperry any tips on how to emulate this?

johnou:
@JoeHegarty been running this in production for a few days now, seems good!

PR #185, opened 2016-08-03
Increase queue limits.

johnou:
1000 is far too low for streams to deal with bursts.


JoeHegarty:
Yeh, I've seen this limit being hit in some of our internal load tests. I'll merge this change for now, but I'm also working on a rearchitect and I'll revisit this case to see if there something more intelligent we can do.

johnou:
Thanks! I have also reached out to the maintainer of jctools to see if mpsc is worth while https://github.com/johnou/orbit/commit/f1a30bb1baebcd0332461d2ab78a7e3e98b1146d#commitcomment-18512047

PR #193, opened 2016-10-07
Expanded ExceptionUtils and added tests

JoeHegarty:
This extends ExceptionUtils by adding a getCauseInChain method which is useful in some cases.
It also cleans up the class a little and adds a test covering ExceptionUtils which didn't exist until today.

### DIFF
--- a/commons/src/main/java/cloud/orbit/util/ExceptionUtils.java
+++ b/commons/src/main/java/cloud/orbit/util/ExceptionUtils.java
@@ -46,6 +46,7 @@ public class ExceptionUtils
         return false;
     }
 
+    @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


JoeHegarty:
Reviewed internally at EA and approved.

commons/src/main/java/cloud/orbit/util/ExceptionUtils.java
```diff
@@ -45,4 +45,17 @@ public static boolean isCauseInChain(Class<?> cause, Throwable chain)
         }
         return false;
     }
+
+    @SuppressWarnings("unchecked")
+    public static <T extends Throwable> T getCauseInChain(final Class<T> cause, final Throwable chain)
+    {
+        for(Throwable ex = chain; ex != null; ex = ex.getCause())
```
johnou: Depth check for possible infinite recursion?

PR #198, opened 2016-10-25
Cleanup stream state if there are no subscribers.

johnou:


johnou:
Failed tests:   testStreamPersistence(cloud.orbit.actors.test.streams.StreamPersistenceTest): expected:<hello3> but was:<null>
  testStreamPersistence(cloud.orbit.actors.test.streams.StreamTest): expected:<hello2> but was:<null>

JoeHegarty:
Do you know why this is failing?

johnou:
Not yet.

johnou:
When shutting down a node the checkAlive fails (because the node will no longer respond to requests) and the subscriber is removed. I think clearing the state without this check is a huge improvement to prevent state storage leaks. We can revisit shutdown another time.

johnou:
leak

JoeHegarty:
Looks good. Thanks Johno

PR #203, opened 2016-11-10
Split out Invocation and allow adding hooks

JoeHegarty:


johnou:
These commits completely break Execution, InvocationHandler must be stateless (in the sense that it cannot hold references to executions in fields). I will create a follow up merge request with a fix.

johnou:
See https://github.com/orbit/orbit/pull/205

PR #207, opened 2016-11-13
Correct usage of await (onInvocation return type cannot be void..).

johnou:


actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
```diff
@@ -603,28 +600,28 @@ public void onRead(final HandlerContext ctx, final Object msg) throws Exception
         }
     }
 
-    private void onInvocation(final HandlerContext ctx, final Invocation invocation)
+    private Task onInvocation(final HandlerContext ctx, final Invocation invocation)
```
johnou: Without this onInvocation essentially becomes a blocking invocation when looking up the cached address.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
```diff
@@ -391,12 +389,11 @@ public void actorDeactivated(RemoteReference remoteReference)
                 nodeAddress = otherNodeAddress;
             }
 
-            // Add to local cache
-            setCachedAddress(actorReference, Task.fromValue(nodeAddress));
```
johnou: redundant and probably break chaining.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
```diff
@@ -283,7 +283,7 @@ private void updateServerNodes()
     private Task<NodeAddress> locateActiveActor(final RemoteReference<?> actorReference)
     {
         final NodeAddress address = await(getCachedAddressTask(actorReference));
-        if (address != null && activeNodes.containsKey(address))
+        if (address != null && address != nullAddress && activeNodes.containsKey(address))
```
johnou: Special short circuit to avoid map lookup.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
johnou: (await invoked from a method not returning completable future will simply invoke get / join.

PR #208, opened 2016-11-15
Check system property to use different jgroups.xml config file

crisperson:
I added a line to check whether there is a system property override to choose the location of the jgroups.xml file Orbit uses in JGroupsClusterPeer. We have a case where we need to change the jgroups.xml for non-prod and prod environments, but would like to keep the container/artifact the same as it's promoted through environments.


JoeHegarty:
As discussed offline, we should probably use an Addon as you are using orbit-hk2 for your project.
For regular projects, this can be achieved by doing something like.

```java
JGroupsClusterPeer clusterPeer = new JGroupsClusterPeer();
clusterPeer.setJgroupsConfig(System.getProperty("orbit.jgroups.config", "classpath:/jgroups.xml"));
stage.setClusterPeer(clusterPeer);
```

PR #209, opened 2016-12-15
added metric extension

vanjoeyg:


JoeHegarty:
Hi Joey. This looks good.
As discussed, we'd like this to be a separate extension. Please submit a PR to https://github.com/orbit/orbit-dropwizard-metrics with it.

Thanks,
Joe

actors/core/pom.xml
```diff
@@ -130,5 +134,34 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-ganglia</artifactId>
+            <version>${dropwizard.metrics.version}</version>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-graphite</artifactId>
```
bilal: Do we need all three reporters? Ganglia, Graphite and DataDog?

JoeHegarty: Currently, it doesn't look like you can actually pass the config in. But if we allowed the config to be passed as the constructor of the MetricsExtension we could assume the user will provide the reporter (though we may want to provide some defaults)

PR #220, opened 2017-02-19
Fixes #218

### LINKED ISSUE #218: Conflicting documentation with respect to timers
The wiki page for timers states:
"Timers do not keep an actor activated and will disappear on deactivation."

However, the method comment for AbstractActor.registerTimer reads:
"The timer is automatically disposed on deactivation.  The timer calls will keep the actor active."

Do timers keep an actor active or not?

johnou:


actors/core/src/main/java/cloud/orbit/actors/runtime/AbstractActor.java
```diff
@@ -240,8 +240,6 @@ protected String runtimeIdentity()
      *
      * @return unique identity string
      */
-
-    @Deprecated
```
johnou: hope you don't mind, we use this all over our project and it seems much easier to access the ref id string directly instead of unwrapping the id through 3-4 method calls.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
```diff
@@ -466,7 +466,10 @@ private RemoteKey createRemoteKey(final RemoteReference actorReference)
                 {
                     return null;
                 }
-                logger.warn("No node available at the moment to place actor: {}.", interfaceClassName);
+                if (logger.isDebugEnabled())
```
johnou: cherry-picked this so it shouldn't cause merge conflict with the other pull request.

PR #221, opened 2017-02-20
Actually wait for servers

### LINKED ISSUE #217: Hosting busy loop
Hosting can go into a crazy busy loop (most noticeable when the node is shutdown), there was a recent change that added logging to this method which completely kills the node. During _one_ test it logged about 3k lines.. [logging.txt](https://github.com/orbit/orbit/files/781042/logging.txt)

johnou:
Motivation:

When Hosting selects a node and no server is available it should wait until another server joins the cluster or wait a few seconds to avoid burning CPU cycles.

Modifications:

Increase wait timeout from 5ms to 2500ms.

Result:

Fixes #217

### DIFF
--- a/actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/reentrant/ReentrantTest.java
+++ b/actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/reentrant/ReentrantTest.java
@@ -58,8 +58,6 @@ public class ReentrantTest extends ActorBaseTest
         Task<Void> doSomething2(String goStr);
 
         Task<Void> doSomething3();
-
-        Task<Void> doSomething4();
     }
 
     public static class ReentrantActorImpl extends AbstractActor implements ReentrantActor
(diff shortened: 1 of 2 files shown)


actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/reentrant/ReentrantTest.java
```diff
@@ -163,22 +154,4 @@ public void testMethodsRunningInParallel() throws Exception
...
-
-    @Test(timeout = 10000)
-    public void testReentrantChain() throws Exception
-    {
-        Stage stage1 = createStage();
-        try
-        {
-            ReentrantActor actor = Actor.getReference(ReentrantActor.class, "1");
-            Task task1 = actor.doSomething4().thenCompose(x -> {
-                return actor.doSomething3();
```
johnou: finally removed broken test.

PR #223, opened 2017-03-04
Cleanup invocation handler.

johnou:
SingleNodeBenchmark.requestThroughput performance is the same, class is much cleaner now.


johnou:
InvocationHandler may pass through null method now but that's correct behaviour.

PR #225, opened 2017-03-22
Feature/extensible message serializer

cgander:
We needed to extend the existing JavaMessageSerializer to customize the way the object replacement works.   I extracted some of the anonymous classes into public ones so that we could extend both the JavaMessageSerializer itself, and the InputStream and OutputStream as well.   The implementations are not changed.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Execution.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Execution.java
@@ -168,9 +168,9 @@ public class Execution extends AbstractExecution implements Startable
         }
         catch (Throwable exception)
         {
-            if (logger.isDebugEnabled())
+            if (logger.isErrorEnabled())
             {
-                logger.debug("Unknown application error. ", exception);
+                logger.error("Unknown application error. ", exception);
             }
             if (invocation.getCompletion() != null)
             {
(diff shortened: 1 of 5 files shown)


johnou:
@cgander I added Kyro serializer to our project which boasts 1000% improvement in speed / size, happy to provide the implementation but I don't have time to integrate it back at the moment.

johnou:
@cgander @JoeHegarty https://gist.github.com/johnou/468643567261e682190657f76d1e7d5b

johnou:
http://java-persistence-performance.blogspot.it/2013/08/optimizing-java-serialization-java-vs.html

johnou:
```java
       KryoSerializer kryoSerializer = new KryoSerializer();
        builder.objectCloner(kryoSerializer);
        builder.messageLoopbackObjectCloner(kryoSerializer);

        stage = builder.build();
        stage.setMessageSerializer(kryoSerializer);
```

johnou:
cc @ceckhardt

johnou:
while there is a dependency on Hibernate I would appreciate it if you could leave support for that by using the same approach as this serializer https://github.com/magro/memcached-session-manager/blob/master/kryo-serializer/src/main/java/de/javakaffee/web/msm/serializer/kryo/HibernateCollectionsSerializerFactory.java

johnou:
I updated https://gist.github.com/johnou/468643567261e682190657f76d1e7d5b which should be a pretty good base, we are happy to maintain our own serializer for Hibernate, forget my previous comment.

PR #227, opened 2017-04-05
Extract and restructure InvocationHandler for easier extension.

johnou:
Not sure if anyone has extended these classes but it probably makes sense to bump to 0.10.0.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/MessageLoopback.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/MessageLoopback.java
@@ -75,7 +75,7 @@ public class MessageLoopback extends NamedPipelineExtension
                     {
                         clonedPayload = cloner.clone(originalPayload);
                     }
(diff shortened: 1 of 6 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/runtime/MessageLoopback.java
```diff
@@ -75,7 +75,7 @@ public Task write(final HandlerContext ctx, final Object msg) throws Exception
                     {
                         clonedPayload = cloner.clone(originalPayload);
                     }
-                    catch (Exception e)
+                    catch (LinkageError | Exception e)
```
johnou: classdefnotfound is possible, didn't change this to throwable because it's not good practice to catch OOM.

johnou: other option is catching error and throwable, and rethrowing error immediately, wdyt?

PR #229, opened 2017-04-12
Failing test when attempting to run parallel/detached execution in an actor

jpennell:
My team is under the impression that the following should work without issue, and not break Orbit's guarantee that executions can't run in parallel on the same actor.

```
@Override
public Task<Void> foo() {
    someOtherActor.getId().thenAccept(id-> state().setId(id));
    return Task.done();
}
```

ie. invoking a method on another actor, and in a continuation function modifying the state of the current actor. Note that the above example does not chain and return the task, it's detached.

I added a test to prove that this works as we believed, but the test fails.

In the `ParallelTest` I added in this PR, `testDetached` causes a `ConcurrentModificationException`.

I'm wondering if we were correct in assuming that this is supported?

Thanks!

### DIFF
--- a/actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/ParallelTest.java
+++ b/actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/ParallelTest.java
@@ -1,0 +1,56 @@
+/*
+ Copyright (C) 2017 Electronic Arts Inc.  All rights reserved.
+
+ Redistribution and use in source and binary forms, with or without
+ modification, are permitted provided that the following conditions
+ are met:
+
+ 1.  Redistributions of source code must retain the above copyright
+     notice, this list of conditions and the following disclaimer.
+ 2.  Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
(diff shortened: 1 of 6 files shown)


JoeHegarty:
The behaviour you are seeing is correct, because you are effectively forking the execution.
In order to guarantee the execution works as expected you need to make sure the entire task chain happens within the actors execution context.

```java
@Override
public Task<Void> foo() {
    return someOtherActor.getId().thenAccept(id-> state().setId(id));  
}
```

This will do what you expect, the key difference is that you are returning the first task so it becomes part of the actor execution context. 
You could also wrap an await around the call and continue to have your Task.done().

The above code likely doesn't compile because of conflicting types but it should give you the right idea.

Does this make sense?

jpennell:
Makes perfect sense. 

Thanks for the clarification Joe.

JoeHegarty:
One other thing, if you want other code to be able to run while that is happening you can always use the Reentrant annotation. The effect here is that code can "interleave", in the above example between the getId and when the thenAccept runs.

You need to make sure you code will handle that correctly, but it can improve performance. 
You annotate any methods that are safe to be interleaved with Reentrant.

See https://github.com/orbit/orbit/wiki/Concepts%3A-Reentrancy

PR #230, opened 2017-04-26
Async API for ClusterPeer.

johnou:


johnou:
I'll split this into multiple pull requests.

PR #236, opened 2017-05-06
Clear local cache of nodes which left the cluster, shutdown reminders…

johnou:
… on shutdown, set local node mode and decrease deactivation timeout default to something sensible.


johnou:
For some reason cloud.orbit.actors.Stage#stopActors doesn't wait for the Task returned by cleanupActors..

PR #237, opened 2017-05-10
Actor Deactivation Flow Changes

JoeHegarty:
This factors out deactivation from Stage and makes it fully asynchronous and lazy.
This is a major architecture change to how deactivation and shutdown works.

Rationale: 
* Makes it easier to reason about the deactivation flow during pulses.
  * A pulse cleanup just starts the deactivation tasks and does not wait on them.
  * Number of concurrent deactivations is handled at the threading level.
  * Ensures only one deactivation for an actor can be running at any one time.
    * This was a potential bug with the existing implementation.
* Allows us to handle shutdowns uniquely as this has always been an area of difficulty for us.
  * This first pass broadly does what the existing shutdown flow did.
* Makes it possible to extend the deactivation rules/system
  * I intend to follow up with a change to expose this via an extension
  * Plan to provide some default deactivation extensions such as culling LRU actors when short on memory.

### DIFF
--- a/actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/ActorBaseTest.java
+++ b/actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/ActorBaseTest.java
@@ -493,7 +493,6 @@ public class ActorBaseTest
             isBusy = false;
             for(Stage stage : stages)
             {
-                System.out.println(stage);
                 if(stage.getLocalObjectsCleaner().hasBacklog())
                 {
                     isBusy = true;
--- a/actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/ActorBaseTest.java
+++ b/actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/ActorBaseTest.java
@@ -486,6 +486,24 @@ public class ActorBaseTest
 
     }
 
+    protected void waitForDeactivations() {
+        boolean isBusy;
+        do
+        {
+            isBusy = false;
+            for(Stage stage : stages)
+            {
+                System.out.println(stage);
+                if(stage.getLocalObjectsCleaner().hasBacklog())
+                {
+                    isBusy = true;
+                    break;
+                }
+
+            }
+        } while(isBusy);
+    }
+
 
     @After
     public void tearDown()
--- a/actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
+++ b/actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
@@ -63,6 +63,7 @@ public class DeactivationTest extends ClientTest
         final UUID id = actor1.getUniqueActivationId().join();
         clock.incrementTime(20, TimeUnit.MINUTES);
         stage.cleanup().join();
+        waitForDeactivations();
         Assert.assertNotEquals(id, actor1.getUniqueActivationId().join());
         dumpMessages();
     }
@@ -78,6 +79,7 @@ public class DeactivationTest extends ClientTest
         final UUID id = actor1.getUniqueActivationId().join();
         clock.incrementTime(20, TimeUnit.MINUTES);
         stage.cleanup().join();
+        waitForDeactivations();
         Assert.assertNotEquals(id, actor1.getUniqueActivationId().join());
         dumpMessages();
     }
@@ -108,6 +110,7 @@ public class DeactivationTest extends ClientTest
         waitFor(() -> isIdle(stage));
         clock.incrementTime(20, TimeUnit.MINUTES);
         stage.cleanup().join();
+        waitForDeactivations();
         client.bind();
         set.add(actor1.getUniqueActivationId().join());
         assertEquals(2, set.size());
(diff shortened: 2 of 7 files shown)


JoeHegarty:
@johnou Can you review this please?

victorhurdugaci:
@JoeHegarty I looked over the PR and didn't see anything obviously wrong, but I'm not very familiar with the internal logic either.

actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/ActorBaseTest.java
```diff
@@ -486,6 +486,24 @@ protected void eventuallyTrue(long timeoutMillis, final Callable<Boolean> callab
...
     }
 
+    protected void waitForDeactivations() {
+        boolean isBusy;
+        do
+        {
+            isBusy = false;
+            for(Stage stage : stages)
+            {
+                System.out.println(stage);
```
johnou: forgot to remove some debug print?

actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/ActorBaseTest.java
```diff
@@ -486,6 +486,24 @@ protected void eventuallyTrue(long timeoutMillis, final Callable<Boolean> callab
...
+                if(stage.getLocalObjectsCleaner().hasBacklog())
+                {
+                    isBusy = true;
+                    break;
+                }
+
+            }
+        } while(isBusy);
```
johnou: I guess this is going to peg a cpu core?

actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
```diff
@@ -63,6 +63,7 @@ public void simpleCleanupTest() throws ExecutionException, InterruptedException
         final UUID id = actor1.getUniqueActivationId().join();
         clock.incrementTime(20, TimeUnit.MINUTES);
         stage.cleanup().join();
+        waitForDeactivations();
```
johnou: why do we need this now? shouldn't it be handled in stage.stop?

actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/ActorBaseTest.java
JoeHegarty: Yep, I'll remove.

actors/test/actor-tests/src/main/java/cloud/orbit/actors/test/ActorBaseTest.java
JoeHegarty: Yes, it's just in tests though. Cleanup now drops straight through so you can't expect any deactivations to actually have taken place.

actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
JoeHegarty: Cleanup drops through now and various tests failed. For simplicity anywhere stage.cleanup is used I added a waitForDeactivations as that gives the same behavior for the tests.
In a real Orbit application we never rely on cleanup doing something immediately in pulse.

johnou: I disagree, cleanup explicitly invokes await(localObjectsCleaner.cleanup()); and we are joining on the cf returned by cleanup so I suspect there is something wrong with the new impl and waitForDeactivations feels like a hack.

JoeHegarty:
The new LocalObjectCleaner cf just represents the submission of the deactivation request. It does not wait for any of them to be complete

Sent from my iPhone

On May 11, 2017, at 23:07, Johno Crawford <notifications@github.com<mailto:notifications@github.com>> wrote:


@johnou commented on this pull request.

________________________________

In actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java<https://github.com/orbit/orbit/pull/237#discussion_r116158004>:

> @@ -63,6 +63,7 @@ public void simpleCleanupTest() throws ExecutionException, InterruptedException
         final UUID id = actor1.getUniqueActivationId().join();
         clock.incrementTime(20, TimeUnit.MINUTES);
         stage.cleanup().join();
+        waitForDeactivations();


I disagree, cleanup explicitly invokes await(localObjectsCleaner.cleanup()); and we are joining on the cf returned by cleanup so I suspect there is something wrong with the new impl and waitForDeactivations feels like a hack.

actors/test/actor-tests/src/test/java/cloud/orbit/actors/test/DeactivationTest.java
JoeHegarty: The new LocalObjectCleaner cf just represents the submission of the deactivation request. It does not wait for any of them to be complete. 
The old cleanup would await the actual deactivations.

JoeHegarty: This change was a conscious decision to simplify the flow. The only place the behaviour matters is tests because the simulate time and expect the deactivation to have happened upon cleanup completing.

PR #244, opened 2017-05-29
Actor construction extension

JoeHegarty:
**Overview**
This splits out the ability to override actor construction into it's own extension.

**Rationale**
It's reasonable to want more than one lifetime extension but that then means the actor construction that runs is effectively random depending on the ordering of those extensions.
With the new ActorConstructionExtension the expectation is that you only have one.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
@@ -43,6 +43,7 @@ import cloud.orbit.actors.cluster.NodeAddress;
 import cloud.orbit.actors.concurrent.MultiExecutionSerializer;
 import cloud.orbit.actors.concurrent.WaitFreeMultiExecutionSerializer;
 import cloud.orbit.actors.extensions.ActorClassFinder;
+import cloud.orbit.actors.extensions.ActorConstructionExtension;
 import cloud.orbit.actors.extensions.ActorDeactivationExtension;
 import cloud.orbit.actors.extensions.ActorExtension;
 import cloud.orbit.actors.extensions.DefaultLoggerExtension;
@@ -62,6 +63,7 @@ import cloud.orbit.actors.runtime.ActorTaskContext;
 import cloud.orbit.actors.runtime.AsyncStreamReference;
 import cloud.orbit.actors.runtime.BasicRuntime;
 import cloud.orbit.actors.runtime.ClusterHandler;
+import cloud.orbit.actors.runtime.DefaultActorConstructionExtension;
 import cloud.orbit.actors.runtime.DefaultDescriptorFactory;
 import cloud.orbit.actors.runtime.DefaultHandlers;
 import cloud.orbit.actors.runtime.DefaultInvocationHandler;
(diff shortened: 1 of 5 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
```diff
@@ -749,6 +751,17 @@ else if (extension.getAfterHandlerName() != null)
             extensions.add(lifetimeExtension);
         }
 
+        ActorConstructionExtension actorConstructionExtension = extensions.stream()
+                .filter(p -> p instanceof ActorConstructionExtension)
+                .map(p -> (ActorConstructionExtension) p)
+                .findFirst().orElse(null);
```
johnou: Or else new DefaultActorConstructionExtension()

actors/runtime/src/main/java/cloud/orbit/actors/runtime/ActorEntry.java
```diff
@@ -107,7 +108,7 @@ public T getObject()
                 return Task.fromValue(null);
             }
         }
-        final Object newInstance = runtime.getFirstExtension(LifetimeExtension.class).newInstance(concreteClass);
+        final Object newInstance = runtime.getFirstExtension(ActorConstructionExtension.class).newInstance(concreteClass);
```
johnou: Is this cached?

JoeHegarty: No, I was actually thinking about that. But it doesn't change the existing behavior so it should be a separate PR.

actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
JoeHegarty: Fair enough, that makes sense.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultActorConstructionExtension.java
```diff
@@ -0,0 +1,51 @@
...
+ * Created by joeh on 2017-05-29.
+ */
+public class DefaultActorConstructionExtension implements ActorConstructionExtension
+{
+    @Override
+    public <T> T newInstance(final Class<T> concreteClass)
+    {
+        try
+        {
+            return concreteClass.newInstance();
```
johnou: Let's consider objensis

johnou: http://objenesis.org

ShawnTuatara: I would expect that to be a choice of the developer working with Orbit. They could explicitly define an ObjenesisActorConstructionExtension and have a dependency on it. I think it is fair enough for Orbit to have a requirement that a default no-arg constructor is available for the default.

JoeHegarty: I agree with @ShawnTuatara. As a general rule it's a requirement of Orbit that actors have a no-arg constructor. An application which wants to do something different is welcome to, but Orbit should not support that by default.
We've also made significant effort within Orbit to reduce our use of third-party dependencies so adding this without a clear benefit would go against that philosophy.

PR #245, opened 2017-05-30
Allow starting stage after it has been previously stopped

armsnyder:
While writing an integration test I noticed that I couldn't simply stop then start the Stage between tests; I had to kill and rebuild the Stage. Based on a comment on the `Startable` interface, I believe this is a bug.

> It must be possible to call start twice without errors.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
@@ -577,7 +577,7 @@ public class Stage implements Startable, ActorRuntime
         logger.info("Starting Stage...");
         extensions = new ArrayList<>(extensions);
         startCalled = true;
-        if (state != null)
+        if (state != null && state != NodeCapabilities.NodeState.STOPPED)
         {
             throw new IllegalStateException("Can't start the stage at this state. " + this.toString());
         }
@@ -890,6 +890,7 @@ public class Stage implements Startable, ActorRuntime
         try
         {
             timer.cancel();
+            timer = null;
         }
         catch (final Throwable ex)
         {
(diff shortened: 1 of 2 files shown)


JoeHegarty:
The comment in `Startable` is actually referring to the fact it may get called more than once before a stage starts.
However, it should be possible to stop and start a stage.
I'll do a couple of tests on this and merge as long as things look good.

actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
```diff
@@ -577,7 +577,7 @@ public void setDeactivationTimeout(long deactivationTimeoutMs)
         logger.info("Starting Stage...");
         extensions = new ArrayList<>(extensions);
         startCalled = true;
-        if (state != null)
+        if (state != null && state != NodeCapabilities.NodeState.STOPPED)
```
ShawnTuatara: If `STOPPED` is the proper check then it would make sense to set the default value for `state` to `STOPPED`. In that case you can replace this with a check for `if(state != NodeState.STOPPED)`.

armsnyder:
I wrote a test that fails and pushed it so this doesn't get merged accidentally.

Getting this error when restarting the stage:
```
java.util.concurrent.CompletionException: org.infinispan.IllegalLifecycleStateException: ISPN000323: Cache 'distributedDirectory' is in 'TERMINATED' state and so it does not accept new invocations. Either restart it or recreate the cache container.
	at org.infinispan.interceptors.InvocationContextInterceptor.handleAll(InvocationContextInterceptor.java:94)
```

johnou:
@armsnyder because the distributed cache is .. cached in Hosting. See cloud.orbit.actors.runtime.Hosting#getDistributedDirectory

actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
johnou: might also want to introduce a lock to prevent concurrent start / stop invocations.

PR #246, opened 2017-05-30
ClusterPeer should cache Cache if necessary.

johnou:
The default cluster peer Infinispan caches the cache already and in some implementations caching the cache at Hosting is not desired and may break the implementation.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
@@ -78,9 +78,6 @@ public class Hosting implements NodeCapabilities, Startable, PipelineExtension
     // according to the micro benchmarks, a guava cache is much slower than using a ConcurrentHashMap here.
     private final Map<RemoteReference<?>, NodeAddress> localAddressCache = new ConcurrentHashMap<>();
 
-    // don't use RemoteReferences, better to restrict keys to a small set of classes.
-    private volatile ConcurrentMap<RemoteKey, NodeAddress> distributedDirectory;
-
     private long timeToWaitForServersMillis = 30000;
 
     private TreeMap<String, NodeInfo> consistentHashNodeTree = new TreeMap<>();
@@ -400,17 +397,7 @@ public class Hosting implements NodeCapabilities, Startable, PipelineExtension
 
     private ConcurrentMap<RemoteKey, NodeAddress> getDistributedDirectory()
     {
-        if (distributedDirectory == null)
-        {
-            synchronized (this)
-            {
-                if (distributedDirectory == null)
-                {
-                    distributedDirectory = clusterPeer.getCache("distributedDirectory");
-                }
-            }
-        }
-        return distributedDirectory;
+        return clusterPeer.getCache("distributedDirectory");
     }
 
     private RemoteKey createRemoteKey(final RemoteReference actorReference)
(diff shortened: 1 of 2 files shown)


johnou:
The default cluster peer Infinispan caches the cache already and in some implementations caching the cache at Hosting is not desired and may break the implementation.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
```diff
@@ -400,17 +397,7 @@ private void setCachedAddress(final RemoteReference<?> actorReference, final Nod
...
-            synchronized (this)
-            {
-                if (distributedDirectory == null)
-                {
-                    distributedDirectory = clusterPeer.getCache("distributedDirectory");
-                }
-            }
-        }
-        return distributedDirectory;
+        return clusterPeer.getCache("distributedDirectory");
```
JoeHegarty: This is a really hot code path, do you have any concerned about the overhead for Infinispan to grab it?
Just wondering if we really don't want to cache this or if the right solution is to clear it during shutdown.

Thoughts?

johnou: I've managed to work around it by creating a wrapper for Ignite.

```java
class ConfigurableIgniteMap<K, V> implements ConcurrentMap<K, V> {

    private final Ignite ignite;
    private final CacheConfiguration<K, V> cacheConfiguration;

    ConfigurableIgniteMap(Ignite ignite, CacheConfiguration<K, V> cacheConfiguration) {
        this.ignite = ignite;
        this.cacheConfiguration = cacheConfiguration;
    }

    @Override
    public V putIfAbsent(@NotNull K k, V v) {
        return ignite.<K, V>getOrCreateCache(cacheConfiguration).getAndPutIfAbsent(k, v);
    }

    @Override
    @SuppressWarnings("unchecked")
    public boolean remove(@NotNull Object k, Object v) {
        return ignite.<K, V>getOrCreateCache(cacheConfiguration).remove((K) k, (V) v);
    }

    @Override
    public boolean replace(@NotNull K k, @NotNull V v, @NotNull V v1) {
        return ignite.<K, V>getOrCreateCache(cacheConfiguration).replace(k, v, v1);
    }

    @Override
    public V replace(@NotNull K k, @NotNull V v) {
        return ignite.<K, V>getOrCreateCache(cacheConfiguration).getAndReplace(k, v);
    }

    @Override
    public int size() {
        throw new UnsupportedOperationException("Not supported");
    }

    @Override
    public boolean isEmpty() {
        throw new UnsupportedOperationException("Not supported");
    }

    @Override
    public boolean containsKey(Object o) {
        throw new UnsupportedOperationException("Not supported");
    }

    @Override
    public boolean containsValue(Object o) {
        throw new UnsupportedOperationException("Not supported");
    }

    @Override
    @SuppressWarnings("unchecked")
    public V get(Object o) {
        return ignite.<K, V>getOrCreateCache(cacheConfiguration).get((K) o);
    }

    @Override
    public V put(K k, V v) {
        return ignite.<K, V>getOrCreateCache(cacheConfiguration).getAndPut(k, v);
    }

    @Override
    @SuppressWarnings("unchecked")
    public V remove(Object o) {
        return ignite.<K, V>getOrCreateCache(cacheConfiguration).getAndRemove((K) o);
    }

    @Override
    public void putAll(@NotNull Map<? extends K, ? extends V> map) {
        ignite.<K, V>getOrCreateCache(cacheConfiguration).putAll(map);
    }

    @Override
    public void clear() {
        ignite.<K, V>getOrCreateCache(cacheConfiguration).clear();
    }

    @NotNull
    @Override
    public Set<K> keySet() {
        throw new UnsupportedOperationException("Not supported");
    }

    @NotNull
    @Override
    public Collection<V> values() {
        throw new UnsupportedOperationException("Not supported");
    }

    @NotNull
    @Override
    public Set<Entry<K, V>> entrySet() {
        throw new UnsupportedOperationException("Not supported");
    }
}
```

PR #247, opened 2017-05-30
Kryo message serializer 

JoeHegarty:
This is the Kryo message serializer contributed by Johno from Sulake.

It serves as both a message serializer and an object cloner so replaces the existing KryoCloner we have. The default Orbit serializer remains the same so there is no impact on applications unless they choose to use this serializer.

My tests show performance varies between the serializers depending on the workload so it is up to applications to determine which is best in their use case.


actors/runtime/src/main/java/cloud/orbit/actors/runtime/ActorProfiler.java
```diff
@@ -44,7 +43,7 @@
 {
     private final HashMap<Object, ProfilerData> collectedData = new HashMap<>();
     private final ReadWriteLock rwlock = new ReentrantReadWriteLock();
-    private final KryoCloner kryoCloner = new KryoCloner();
+    private final KryoSerializer kryoCloner = new KryoSerializer();
```
ShawnTuatara: Should rename this field to be "serializer" so it describes the capability needed and not the specific implementation.

ShawnTuatara: Why isn't this a MessageSerializer or ObjectInstanceCloner for the field type so if we swap out the implementation we don't change the type?

JoeHegarty: Looks like the ActorProfiler is pretty old from 2015 and and isn't actually used anywhere. It can probably be removed, we have better ways to track execution time of tasks now with invocation handler extensions etc

JoeHegarty: Oh, my apologies. This is only used in the Orbit benchmarks so we do still need it, but otherwise it's not used at runtime in a real application.

JoeHegarty: This class actually just uses this to clone objects for it's own profiling, so it's not actually a generic field for any serializer/cloner, it's specifically for this profiler.

PR #248, opened 2017-05-31
Optimise hosting select node.

johnou:
Looks good in my load tests.


johnou:
cc @ceckhardt @alexj212

PR #249, opened 2017-06-03
InterceptableActorExtension

armsnyder:
Adds the abstract `InterceptableActorExtension` class which decorates `ActorExtension` and allows interception of calls to its delegate ActorExtension.

Example:
```java
StorageExtension existingExtension = getStorageExtensionFromStage(stage);

InterceptableStorageExtension interceptableExtension = 
        new InterceptableStorageExtension(existingExtension);
interceptableExtension.addInterceptor(new TimerStorageInterceptor());

stage.getExtensions().remove(existingExtension);
stage.addExtension(interceptableExtension);
```
```java
class TimerStorageInterceptor implements StorageInterceptor {

    long timeStart;

    @Override
    public Task<Void> preWriteState(final InterceptedValue<RemoteReference<?>> reference,
                                    final InterceptedValue state) {
        timeStart = System.currentTimeMillis();
        return Task.done();
    }

    @Override
    public Task<Void> postWriteState(final InterceptedValue<RemoteReference<?>> reference,
                                     final InterceptedValue state)
    {
        long timeElapsed = System.currentTimeMillis() - timeStart;
        System.out.println("Time elapsed: " + timeElapsed);
        return Task.done();
    }
}
```

Since the `InterceptableActorExtension` is a decorator, it's identical to the inner extension from Stage's point of view. This feature enables authors to extend the functionality of existing extensions without knowing their implementations.

To be honest, I'm not confident this feature will be accepted, which is why I didn't write any JavaDoc or tests. There's a good argument that authors would more readily write their own custom wrapping classes for their specific use cases. So absolutely feel free to close this PR if you share that sentiment.

### DIFF
--- a/actors/core/src/main/java/cloud/orbit/actors/extensions/interceptable/InterceptableActorClassFinder.java
+++ b/actors/core/src/main/java/cloud/orbit/actors/extensions/interceptable/InterceptableActorClassFinder.java
@@ -1,0 +1,84 @@
+/*
+ Copyright (C) 2017 Electronic Arts Inc.  All rights reserved.
+
+ Redistribution and use in source and binary forms, with or without
+ modification, are permitted provided that the following conditions
+ are met:
+
+ 1.  Redistributions of source code must retain the above copyright
+     notice, this list of conditions and the following disclaimer.
+ 2.  Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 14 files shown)


JoeHegarty:
I think you are right in that each user should just wrap the extensions how they want. There is no reason you can't use this interceptor method in your own project though.

PR #252, opened 2017-06-14
Use Caffeine for local address cache.

johnou:


actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
```diff
@@ -354,6 +356,8 @@ public void actorDeactivated(RemoteReference remoteReference)
             // Target node still valid?
             if (activeNodes.containsKey(nodeAddress))
             {
+                // Cache locally
```
johnou: @ceckhardt can you pull this optimization into your PR?

ceckhardt: yep

PR #256, opened 2017-06-22
Always nullify actor on deactivation even if it fails so it can be GC…

johnou:
…'d (deactivation cannot run a second time so it doesn't make sense keeping the object in the actor entry).

Simplify DefaultLocalObjectsCleaner codeflow and fix possible race condition with invoking deactivate twice on an actor entry (not a problem but a waste of cycles).

Replace await with simple thenApply to cut down on bytecode of hot method locateAndActivateActor.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/ActorEntry.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/ActorEntry.java
@@ -166,12 +166,21 @@ public class ActorEntry<T extends AbstractActor> extends ActorBaseEntry<T>
             {
                 return Task.done();
             }
-            return executionSerializer.offerJob(key, () -> doDeactivate(), 10000);
+            return executionSerializer.offerJob(key, this::doDeactivate, 10000);
         }
         catch (final Throwable ex)
         {
-            // this should never happen, but deactivate must't fail.
-            ex.printStackTrace();
+            // this should never happen, but deactivate must not fail.
+            try
+            {
+                getLogger().error("Error executing action", ex);
+            }
+            catch (Throwable ex2)
+            {
+                // just to be on the safe side... loggers can fail...
+                ex2.printStackTrace();
+                ex.printStackTrace();
+            }
             return Task.done();
         }
     }
@@ -183,7 +192,6 @@ public class ActorEntry<T extends AbstractActor> extends ActorBaseEntry<T>
             try
             {
                 await(deactivate(getObject()));
-                actor = null;
             }
             catch (final Throwable ex)
             {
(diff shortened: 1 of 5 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/runtime/ActorEntry.java
```diff
@@ -166,11 +166,11 @@ public T getObject()
...
                 return Task.done();
             }
-            return executionSerializer.offerJob(key, () -> doDeactivate(), 10000);
+            return executionSerializer.offerJob(key, this::doDeactivate, 10000);
         }
         catch (final Throwable ex)
         {
-            // this should never happen, but deactivate must't fail.
+            // this should never happen, but deactivate must not fail.
             ex.printStackTrace();
```
ceckhardt: can these exceptions be logged?

johnou: @ceckhardt maybe something like this?

```java
 try {
    logger.error("Error executing action", error);
    } catch (Throwable ex) {
        // just to be on the safe side... loggers can fail...
        ex.printStackTrace();
    }
```

johnou: nb. do not use actor.logger, it may have been set null!

ceckhardt: yeah, we have uncovered some exceptions being hidden by orbit in previous pull requests, so whenever I see this I always feel like we should be sending them to the logger.  I like it.

johnou: ptal

actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
```diff
@@ -376,25 +376,23 @@ public void actorDeactivated(RemoteReference remoteReference)
...
 
-        // Someone else beat us to placement, use that node
-        if (otherNodeAddress != null)
-        {
-            nodeAddress = otherNodeAddress;
-        }
+            // Someone else beat us to placement, use that node
+            if (otherNodeAddress != null)
+            {
+                nodeAddress1 = otherNodeAddress;
```
ShawnTuatara: Can we come up with a better name than nodeAddress1? What is this actually suppose to represent? If we already have nodeAddress earlier in the method it would be nice to understand the distinction.

johnou: updated!

PR #258, opened 2017-06-26
Updated third-party dependencies

JoeHegarty:
This is a general update of third-party dependencies.

### DIFF
--- a/actors/runtime/pom.xml
+++ b/actors/runtime/pom.xml
@@ -122,12 +122,12 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
         <dependency>
             <groupId>com.esotericsoftware</groupId>
             <artifactId>kryo</artifactId>
-            <version>3.0.3</version>
+            <version>4.0.0</version>
         </dependency>
         <dependency>
             <groupId>de.javakaffee</groupId>
             <artifactId>kryo-serializers</artifactId>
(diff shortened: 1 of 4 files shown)


actors/runtime/pom.xml
```diff
@@ -122,12 +122,12 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
         <dependency>
             <groupId>com.esotericsoftware</groupId>
             <artifactId>kryo</artifactId>
-            <version>3.0.3</version>
+            <version>4.0.0</version>
```
ShawnTuatara: Concerns with breaking changes around serialization: https://github.com/EsotericSoftware/kryo/releases

ShawnTuatara:
Would be nice to have context on the need to increase the version for each of the dependencies that are being updated.

actors/runtime/pom.xml
JoeHegarty: Yes, I spotted that, but the only data that potentially lives between a deployment is actor directory and we never guarantee compatibility.

JoeHegarty:
There is no specific context beyond keeping up to date with our major dependencies so we don't lag behind. This is to ensure we're getting any security fixes and general improvements and to ensure that there is no expensive integration down the road.

JoeHegarty:
The reason that triggered me to go do this now is the new ea-async which includes support for DCEVM agent loading.

PR #259, opened 2017-06-27
Response Cache Optimizations

JoeHegarty:
This resolves an issue discovered by @ceckhardt's team which causes thread contention in the Response Cache system. 
This is caused by locking internally in the JDK when creating a new message digest. It appears this is a known issue and has been worked around in other projects such as Guava and the recommendation is to clone the digest.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
@@ -88,7 +88,7 @@ import cloud.orbit.actors.runtime.RandomSelectorExtension;
 import cloud.orbit.actors.runtime.Registration;
 import cloud.orbit.actors.runtime.ReminderController;
 import cloud.orbit.actors.runtime.RemoteReference;
-import cloud.orbit.actors.runtime.ResponseCaching;
+import cloud.orbit.actors.runtime.DefaultResponseCachingExtension;
 import cloud.orbit.actors.runtime.SerializationHandler;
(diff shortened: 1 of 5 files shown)


jpennell:
I'll pull this and re-run our load tests!

actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
```diff
@@ -715,7 +715,7 @@ public void setDeactivationTimeout(long deactivationTimeoutMs)
                 .findFirst()
                 .orElseGet(() ->
                 {
-                    final ResponseCaching responseCaching = new ResponseCaching();
+                    final DefaultResponseCachingExtension responseCaching = new DefaultResponseCachingExtension();
```
ShawnTuatara: Syntax sugar... make the responseCaching of type ResponseCachingExtension (coding to interfaces)?

johnou: +1

commons/src/main/java/cloud/orbit/concurrent/MessageDigestFactory.java
```diff
@@ -0,0 +1,91 @@
...
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * Created by joeh on 2017-06-27.
+ */
+public class MessageDigestFactory
+{
+    private final String algorithm;
+    private final MessageDigest digestPrototype;
+    private final Boolean supportsClone;
```
johnou: use boolean primitive?

PR #267, opened 2017-07-17
Seems way to high for the default settings of Orbit, thread size in 6…

johnou:
…4bit JVM is 1024k, multiplied by 128..


actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
```diff
@@ -136,7 +136,7 @@
 {
     private Logger logger = LoggerFactory.getLogger(Stage.class);
 
-    private static final int DEFAULT_EXECUTION_POOL_SIZE = 128;
+    private static final int DEFAULT_EXECUTION_POOL_SIZE = 32;
```
johnou: If someone needs 128 threads they should probably be setting that when creating the Stage..

PR #270, opened 2017-08-14
Remove uberjar dependency

### LINKED ISSUE #269: Avoid using Infinispan Uberjars
Since this causes several classpath issues like https://developer.jboss.org/thread/275777.

Instead of depending on ```infinispan-embedded```, please use ```infinispan-core```

gustavonalle:
Fixes #269


johnou:
@JoeHegarty lgtm, we should check the sample project after this change (ensure it has it's own slf4j logger eg. log4j2).

johnou:
@gustavonalle thanks!

PR #276, opened 2017-10-10
Optimize object allocation when using readObjectOrNull and readClassA…

johnou:
…ndObject

Motivation:

yourkit shows there is a high object allocation when using readObjectOrNull and readClassAndObject, we can replace it with int / kv of strings for headers which shows much better performance.

Modifications:

Replace readObjectOrNull and readClassAndObject with simple type checks and serialization methods if possible.

Result:

Higher throughput.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/KryoSerializer.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/KryoSerializer.java
@@ -67,6 +67,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.GregorianCalendar;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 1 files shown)


johnou:
```text
Benchmark                                                     Mode  Cnt       Score      Error  Units
CustomKryoSerializerTest.testSerializeDeserializeHeadersNew  thrpt  100  620239.983 ± 6840.181  ops/s
CustomKryoSerializerTest.testSerializeDeserializeHeadersOld  thrpt  100  302015.556 ± 4107.025  ops/s
```

johnou:
!!! This changes the wire format of messaging (not compatible with previous versions of Orbit).

actors/runtime/src/main/java/cloud/orbit/actors/runtime/KryoSerializer.java
```diff
@@ -376,8 +469,50 @@ public void serializeMessage(BasicRuntime basicRuntime, OutputStream outputStrea
...
+                    kryo.writeClassAndObject(out, message.getObjectId());
+                }
+
+                Map<Object, Object> headers = message.getHeaders();
+                if (headers != null)
+                {
+                    out.writeInt(headers.size());
+                    for (Map.Entry<Object, Object> entry : headers.entrySet())
+                    {
+                        out.writeString(String.valueOf(entry.getKey()));
```
johnou: @JoeHegarty this also means that the key for sticky headers must be string, we might want to do some type checks or the same value type magic as per below, wdyt?

PR #278, opened 2017-10-11
Make Kryo default message serializer

JoeHegarty:
This change makes Kryo the default message serializer in place of the Java message serializer.
Throughput for the Kryo serializer is on average 5x the Java message serializer. 

This is in a large part due to the optimization work @johnou has done around kryo. There are a few cases where the Kryo serializer is less flexible but the performance gain now seems worth the tradeoff. As such, some projects may need to switch back to the Java serializer or make some changes to their DTOs.

```
Kryo: 636254.682 ops/s
Java: 126572.244 ops/s
```

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
@@ -711,7 +711,7 @@ public class Stage implements Startable, ActorRuntime
         }
         if (messageSerializer == null)
         {
-            messageSerializer = new JavaMessageSerializer();
+            messageSerializer = new KryoSerializer();
         }
         if (clusterPeer == null)
         {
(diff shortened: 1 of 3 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/Stage.java
```diff
@@ -711,7 +711,7 @@ public void setBroadcastActorDeactivations(boolean broadcastActorDeactivation)
         }
         if (messageSerializer == null)
         {
-            messageSerializer = new JavaMessageSerializer();
+            messageSerializer = new KryoSerializer();
```
johnou: Can we use a shared kryo serializer for messaging and object cloning?

JoeHegarty: We don't do this for the loopback cloner. Do you see a big advantage and sharing across all 3?

johnou: Lower memory footprint.

JoeHegarty:
The test failures seem to be timing related as they all involve streams or observers and the selection seems random. My feeling is these are bad tests that need some kind of wait. Hopefully myself or @johnou can look into these shortly

PR #279, opened 2017-10-26
Support for skipping last access time when invoking actors.

johnou:


johnou:
Unfortunately another actor implementation annotation, I didn't want to have to add a boolean to Invocation AND Message which would travel over the wire just for this feature. What we could do in the future is improve cloud.orbit.util.AnnotationCache#isAnnotated to transverse method.getDeclaringClass().getInterfaces().. but I think that's out of scope for this cs.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/ActorEntry.java
```diff
@@ -296,13 +301,4 @@ public synchronized void clearTimers()
         }
         return Task.done();
     }
-
-    @Override
-    public String toString()
```
johnou: in the end it was nicer to use the default toString implementation as it caused a stackoverflow for stateless actors.

JoeHegarty:
Overall I think this looks good.
I'd like to land it with https://github.com/orbit/orbit/pull/280 as they both deal with similar kinds of requirements

PR #280, opened 2017-11-03
Actor Time to Live Override Annotation

JoeHegarty:
This change allows Orbit users to override the time to live of an actor type.

Motivation: Currently it is only possible to set a global time to live. It is an often requested feature that this should be tuneable per actor type.

Example:
```java
@TimeToLive(value=10, timeUnit = TimeUnit.SECONDS)
public interface User extends Actor{}
```

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultLocalObjectsCleaner.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultLocalObjectsCleaner.java
@@ -32,6 +32,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import cloud.orbit.actors.annotation.NeverDeactivate;
+import cloud.orbit.actors.annotation.TimeToLive;
 import cloud.orbit.actors.concurrent.ConcurrentExecutionQueue;
 import cloud.orbit.actors.extensions.ActorDeactivationExtension;
 import cloud.orbit.concurrent.Task;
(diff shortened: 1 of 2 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultLocalObjectsCleaner.java
```diff
@@ -166,16 +167,39 @@ private Task cleanupActors(final boolean shutdownAll)
...
         {
             return false;
         }
-        // Check for timeout
-        boolean shouldRemove = clock.millis() - actorEntry.getLastAccess() > defaultActorTTL;
+
+        // Check for ttl override
+        if(interfaceClass.isAnnotationPresent(TimeToLive.class))
+        {
+            final TimeToLive customTtl = interfaceClass.getAnnotation(TimeToLive.class);
```
johnou: Better to use our annotation cache here.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultLocalObjectsCleaner.java
```diff
@@ -166,16 +167,39 @@ private Task cleanupActors(final boolean shutdownAll)
 
     private boolean shouldRemove(final ActorBaseEntry<?> actorEntry, final Set<ActorBaseEntry<?>> toRemove)
     {
+        final Class<?> interfaceClass = RemoteReference.getInterfaceClass(actorEntry.getRemoteReference());
         // Make sure it isn't tagged NeverDeactivate
-        if (RemoteReference.getInterfaceClass(actorEntry.getRemoteReference()).isAnnotationPresent(NeverDeactivate.class))
+        if (interfaceClass.isAnnotationPresent(NeverDeactivate.class))
```
johnou: And here.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultLocalObjectsCleaner.java
JoeHegarty: I thought we decided we didn't need to use the annotation cache in these cases because the JVM cached? The annotation cache was for cases when we were inspecting methods, or am I misremebering?

johnou: Think you're right.

PR #281, opened 2017-11-05
Custom timeout annotation for actor methods.

johnou:


johnou:
@quilombodigital ptal.

johnou:
One thing to note of course is that clean up runs every five seconds.

johnou:
Small note, I find it confusing that Message has a method field that's null on the receiving node.

JoeHegarty:
Yeh, the method field thing is confusing, it has tripped me up a couple of times.
For Orbit 2 we should look at mutating the message through the pipeline so it's strictly typed at each step.

actors/core/src/main/java/cloud/orbit/actors/annotation/Timeout.java
```diff
@@ -0,0 +1,49 @@
...
+ * Defines a custom timeout for the given actor method that will be used instead of the
+ * configured <code>responseTimeoutMillis</code>.
+ *
+ * @author Johno Crawford (johno@sulake.com)
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Timeout
+{
+    int ttlDuration();
```
JoeHegarty: Should this be "value" so you don't need to specify the name if you want to use the default unit? That seems to be the commonly agreed convention.

actors/core/src/main/java/cloud/orbit/actors/annotation/Timeout.java
```diff
@@ -0,0 +1,49 @@
...
+    TimeUnit ttlUnit() default TimeUnit.SECONDS;
```
JoeHegarty: Should this be timeout instead of ttl?

quilombodigital: If I can give my 2 cents... from my perspective, a "timeout" is used when you have a reaction, "throws an exception". "TTL" I think naturally means something just expired and it will be discarded.

JoeHegarty: I think we should just go with `value` and `timeUnit` like the new `TimeToLive` annotation, that way we avoid the naming question entirely.

quilombodigital: By the way, I dont know if this is already being done, but I think it would be a good idea to store the discarded count (by class/method?) in a metric, So we can see exactly what is happening when the system is under pressure. Just because there are messages that can be discarded, this does not mean we dont need to know when this is happening. :)

JoeHegarty: This will just fail like it does today, so your Task will contain a timeout exception instead of a result. What you do with that is up to you.

johnou: Renamed to the same as TimeToLive annotation.

PR #282, opened 2017-11-10
Kryo serializer improvements.

johnou:
```
Benchmark                                                Mode  Cnt       Score      Error  Units
ExtendedKryoSerializerTest.testSerializeDeserializeNew  thrpt  100  624227.795 ± 2637.346  ops/s
ExtendedKryoSerializerTest.testSerializeDeserializeOld  thrpt  100  594780.377 ± 3490.899  ops/s
```


johnou:
After these changes older versions of Orbit are not wire compatible.

PR #283, opened 2017-11-20
On-Demand actor deactivation

BogdanEAM:
Implement a way to request actor deactivation by calling a method annotated with @Deactivate

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Execution.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Execution.java
@@ -82,8 +82,17 @@ public class Execution extends AbstractExecution implements Startable
         final LocalObjects.LocalObjectEntry<Object> entry = objects.findLocalObjectByReference(toReference);
         if (entry != null)
         {
-            final Task<Object> result = InternalUtils.safeInvoke(() -> entry.run(
-                    target -> performInvocation(ctx, invocation, entry, target)));
+            Task<Object> result;
+
+            if (invocation.isDeactivate())
+            {
+                result = runtime.deactivateActor(entry);
+            }
+            else
+            {
+                result = InternalUtils.safeInvoke(() -> entry.run(
+                        target -> performInvocation(ctx, invocation, entry, target)));
+            }
             // this has to be done here because of exceptions that can occur before performInvocation is even called.
             if (invocation.getCompletion() != null)
             {
@@ -92,7 +101,7 @@ public class Execution extends AbstractExecution implements Startable
         }
         else
         {
-            if (toReference instanceof Actor)
+            if (toReference instanceof Actor && !invocation.isDeactivate())
             {
                 // on activate will handle the completion;
                 InternalUtils.safeInvoke(() -> executionSerializer.offerJob(toReference,
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Execution.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Execution.java
@@ -86,7 +86,10 @@ public class Execution extends AbstractExecution implements Startable
 
             if (invocation.isDeactivate())
             {
-                result = runtime.deactivateActor(entry);
+                result = InternalUtils.safeInvoke(() -> entry.run(
+                        target -> performInvocation(ctx, invocation, entry, target)));
+
+                runtime.deactivateActor(entry).join();
             }
             else
             {
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Invocation.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Invocation.java
@@ -28,6 +28,7 @@
 
 package cloud.orbit.actors.runtime;
 
+import cloud.orbit.actors.annotation.Deactivate;
 import cloud.orbit.actors.cluster.NodeAddress;
 import cloud.orbit.concurrent.Task;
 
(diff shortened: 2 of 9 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/runtime/Invocation.java
```diff
@@ -48,6 +49,11 @@
     private int hops;
     private int messageId;
 
+    /**
+     * Triggers actir deactivation is set
```
ShawnTuatara: Nit: Spelling and grammar.

ShawnTuatara:
Can we have some context as to why we think `@Deactivate` annotation on the actor is the best approach for handling this? I can see use cases where we want to forcefully deactivate actors as part of gracefully shutting down the stage. I can also see use cases where the Actor itself has determined it should be deactivated and as such should have a way to tell the Stage it wants to deactivate. In the first case I don't see any need to have the actor method and annotation at all. In the second case it is within the Actor to call something to deactivate. This looks like a case where a client calling the node the actor is on is having authority on the life of the actor. Is that the contract we want to expose?

BogdanEAM:
@ShawnTuatara My changes are just for adding a way that a client can request actor deactivation.
The graceful deactivation when stopping the Stage is already implemented, see Stage.stop().
Your second case is the subject of another change that I'm not addressing in this PR and I'm not planning to implement it myself unless it is decided that is mandatory for on-demand Actor deactivation.

ShawnTuatara:
What is the use case where we want the client to have control over the lifecycle of the Actor? Can you give more details / context?

BogdanEAM:
I think the client already has (some) control over the lifecycle of the Actor because it can trigger Actor activation.
An example would be if you want to block/ban an account at Application level you would also want to deactivate all it's actors.

ShawnTuatara:
That is where it is interesting because with virtual actors the client should never know if the actor is active or not. The fact that the client can activate an Actor is not because it explicitly said so but more because it wanted to perform an action against the Actor. It is the Stage that has deemed the Actor should activate as it is needed to have that Actor available to process the action.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/Execution.java
```diff
@@ -82,8 +82,20 @@ protected void onInvocation(final HandlerContext ctx, final Invocation invocatio
         final LocalObjects.LocalObjectEntry<Object> entry = objects.findLocalObjectByReference(toReference);
         if (entry != null)
         {
-            final Task<Object> result = InternalUtils.safeInvoke(() -> entry.run(
-                    target -> performInvocation(ctx, invocation, entry, target)));
+            Task<Object> result;
+
+            if (invocation.isDeactivate())
+            {
```
BogdanEAM: Not sure if the code from this branch works because of the concurrency. Is it possible that the "performInvoke" is actually executed after the runtime.deactivateActor() and to re-activate the actor?

JoeHegarty:
## Philosophical Arguments
Not knowing about lifetime is a key feature of virtual actors, but I don't think offering a way to request deactivation is necessarily an issue. Like `@CacheResponse` and `@OnlyIfActivated` I see it as giving hints to the framework to better optimize your application, and there is great value in allowing experts more control in scenarios where they clearly understand the implications.

For example, if you were building an MMO and you had an actor which represented the user session/connection. When that connection drops it is reasonable for the application to notify the framework that the actor is no longer needed and can be deactivated. 

However, it's key to note that this would never be a guarantee but it would be a "best effort". There are edge cases (such as the message failing to be delivered) which may impact Orbit's ability to deactivate the actor and as a design choice we often say the framework is free to ignore requests of this nature for any reason.

## Current Approach
I apologize for not being able to offer guidance on this before now.
I don't think this PR is the right approach.

The `@Deactivate` annotation changes the semantics of the message significantly. 
Up until now message annotations may cache a response or throw the message away etc but they never fundamentally change the meaning of it.
With this new annotation I can create a message called `start` which actually deactivates the actor and that's not clear from the call site.
The fact that we have to add a flag to `Invocation` is very intrusive and highlights that we're fundamentally altering the meaning of the message.

## Potential Approach
I've been thinking about this overnight and I have a potential approach.
I think the right way is to have an `Observer` on each node listening for deactivation messages. We do something similar with the `ExecutionCacheFlushManager` except that's a broadcast.

For example, you could add a static `requestDeactivation` to the Actor interface similar to `getReference` and then register an observer on each node.

Something like: 

**DeactivateObserver.java**
```java
public interface DeactivateObserver extends ActorObserver {
    Task<Void> deactivateActor(Actor actor)
}
```

**DefaultDeactivateObserver.java**
```java
public class DefaultDeactivateObserver implements DeactivateObserver {
     // Override deactivateActor and do the deactivation
}
```

**Actor.java**
```java
public static Task requestDeactivation(final Actor actor)
{
    NodeAddress address = await(ActorRuntime.getRuntime().locateActor(RemoteReference.from(actor), false));
    if(address != null) {
        return DefaultDescriptorFactory.observerRef(address, DeactivateObserver.class, "").deactivateActor(actor);
    }
}

**Stage.java**
```java
// Somewhere during start
registerObserver(DeactivateObserver.class, "", new DefaultDeactivateObserver());
```

Hopefully that gives you an idea of where to go with this.

johnou:
You might be able to extend cloud.orbit.actors.runtime.DefaultLocalObjectsCleaner#shouldRemove to allow clients to pass in actors to deactivate using an observer.

Issue #292, opened 2018-02-08
Option for disabling message loopback


johnou:
Useful for benchmarking new cluster peers.

PR #293, opened 2018-04-03
Reduce object allocation with MessageSerializer.

johnou:
No pooling

```
Benchmark                                         Mode  Cnt       Score      Error  Units
SingleNodeBenchmark.requestThroughput            thrpt   40  163676.503 ± 1647.612  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt   40       3.958 ±    0.086  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt   40      20.065 ±    0.745  us/op
```

Pooling

```
Benchmark                                         Mode  Cnt       Score      Error  Units
SingleNodeBenchmark.requestThroughput            thrpt   40  170217.244 ± 1900.282  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt   40       3.808 ±    0.053  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt   40      19.840 ±    1.029  us/op
```

Fixes https://github.com/orbit/orbit/issues/292


actors/runtime/src/main/java/cloud/orbit/actors/runtime/KryoSerializer.java
```diff
@@ -483,9 +486,9 @@ public void serializeMessage(BasicRuntime basicRuntime, OutputStream outputStrea
                 writeHeaders(kryo, out, message.getHeaders());
                 writeNodeAddress(out, message.getFromNode());
                 writePayload(kryo, out, message);
-                return null;
-            }
-        });
+                return out.toBytes();
```
johnou: no risk of data corruption here, new byte array is returned, not the pooled byte array.

PR #295, opened 2018-04-04
Add node placement group support

cesarbs:
Placement groups enable zero downtime deployments by spinning up new nodes in a new placement group while configuring nodes in the previous placement group to direct new actor activations to nodes in the the new placement group. Once all or most actors have moved to nodes in the new placement group, the old placement group's nodes can be taken down with no or minimal impact to service availability.

This change adds a placement group property to `NodeCapabilities` to hold a node's placement group, as well as a dynamically configurable collection of target placement groups to `Hosting` so that only nodes in those target groups are selected for actor activations.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
@@ -44,7 +44,6 @@ import cloud.orbit.actors.exceptions.ObserverNotFound;
 import cloud.orbit.actors.extensions.NodeSelectorExtension;
 import cloud.orbit.actors.extensions.PipelineExtension;
 import cloud.orbit.actors.net.HandlerContext;
-import cloud.orbit.annotation.Config;
 import cloud.orbit.concurrent.Task;
 import cloud.orbit.exception.UncheckedException;
 import cloud.orbit.lifecycle.Startable;
@@ -59,6 +58,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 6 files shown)


JoeHegarty:
Thanks @cesarbs.
As discussed on slack. 
* Please remove the check in `canActivate`
* In selectNode you can't await on all the nodes because it effectively grinds the current node to a halt if a remote one is misbehaving/doesn't respond.
  * You should do what we're doing with the canActivate calls lower in the loop and do it lazily.

actors/runtime/src/main/java/cloud/orbit/actors/runtime/Hosting.java
```diff
@@ -433,13 +452,16 @@ private RemoteKey createRemoteKey(final RemoteReference actorReference)
...
             final List<NodeInfo> currentServerNodes = serverNodes;
+            final Set<String> currentTargetPlacementGroups = targetPlacementGroups;
 
             // filter out nodes which cannot host actors (nodes which are not running and nodes which cannot host this actor)
+            // and nodes that are not in any target placement group
             final List<NodeInfo> potentialNodes = currentServerNodes.stream()
                     .filter(n -> (!n.cannotHostActors && n.state == NodeState.RUNNING)
-                            && actorSupported_no != n.canActivate.getOrDefault(interfaceClassName, actorSupported_yes))
+                            && actorSupported_no != n.canActivate.getOrDefault(interfaceClassName, actorSupported_yes)
+                            && (n.placementGroupPending.get() || currentTargetPlacementGroups.contains(n.placementGroup)))
```
JoeHegarty: This should be `(!n.placementGroupPending.get() && currentTargetPlacementGroups.contains(n.placementGroup))`

cesarbs: Oops, yep.

cesarbs: Actually no, not really (at first glance I though this was the filter down below. What I'm doing here is getting nodes that we either haven't read their placement group yet (`placementGroupPending == true`), or we know they're in a target placement group. The ones whose placement group is still unknown will be updated in the loop below.

JoeHegarty: Ok, that makes sense. Then I guess this is all good.

PR #313, opened 2018-12-07
Cache Optimizations

JoeHegarty:
Backport of https://github.com/ceckhardt/orbit/pull/11


actors/runtime/src/main/java/cloud/orbit/actors/runtime/DefaultResponseCachingExtension.java
```diff
@@ -88,15 +103,35 @@ public static void setClock(final Clock clock)
         DefaultResponseCachingExtension.clock = clock;
     }
 
+    public Map<Method, Long> getCacheSizes() {
+
+        return masterCache.asMap().entrySet().stream()
+                .map(this::getCacheSize)
```
cesarbs: This mapping seems unnecessary since you could just do `entry.geyKey()` and `entry.getValue().estimatedSize()` directly in the collector call, but if it's not a hot path I guess it doesn't matter and is slightly more readable.

PR #321, opened 2019-03-11
Don't complete exceptional tasks normally

msillanpaa:
We observed an issue where something was completed exceptionally in an actor but at the caller the return value looked like it had completed normally. That led to checking this part.

I've also signed the CLA to be able to contribute.


johnou:
Nice catch!

msillanpaa:
We observed an issue where something was completed exceptionally in an actor but at the caller the return value looked like it had completed normally. That led to checking this part. 

I've also signed the CLA to be able to contribute.

JoeHegarty:
Available in 1.12.0
https://github.com/orbit/orbit/releases/tag/v1.12.0

PR #351, opened 2019-03-31
DSL: add list and map types

cesarbs:
Addresses #337 

Can now do

```
data MyData {
  list<string> aList;
  map<string, int32> aMap;
}
```

And

```
actor MyActor {
  map<string, int32> aMethod(list<int64> arg);
}
```

Works with user-defined types as well, and can be nested (e.g. `list<map<string, list<int32>>>`).

Very rudimentary support with essentially no validation for now. I'll add validation in a subsequent PR.


src/dsl/orbit-codegen-dsl/src/main/antlr/cloud/orbit/dsl/Orbit.g4
```diff
@@ -20,11 +20,13 @@ enumDeclaration: ENUM name=ID LC_BRACE members=enumMember* RC_BRACE ;
...
-actorMethod: returnType=ID name=ID L_PAREN (args=methodParam (COMMA args=methodParam)*)? R_PAREN SEMI_COLON;
-methodParam: type=ID name=ID ;
+actorMethod: returnType=type name=ID L_PAREN (args=methodParam (COMMA args=methodParam)*)? R_PAREN SEMI_COLON;
+methodParam: type name=ID ;
 
 dataDeclaration: DATA name=ID LC_BRACE fields=dataField* RC_BRACE ;
-dataField: type=ID name=ID EQUAL index=INT SEMI_COLON ;
+dataField: type name=ID EQUAL index=INT SEMI_COLON ;
+
+type: name=ID (L_ANGLE of=type (COMMA of=type)* R_ANGLE)? ;
```
victorhurdugaci: 1. I would add one more level to type: `generic` and `non generic.` It will make extending the grammar easier because there will be cases in which you want just a normal type without supporting generics. 
2. nitpick: consider naming it `typeName` instead of `type`

src/dsl/orbit-codegen-dsl/src/test/kotlin/cloud/orbit/dsl/OrbitFileParserTest.kt
```diff
@@ -143,6 +143,134 @@ class OrbitFileParserTest {
...
+        val actualCu = OrbitFileParser().parse(text, FAKE_PACKAGE_NAME)
+
+        Assertions.assertEquals(expectedCu, actualCu)
+    }
+
+    @Test
+    fun parseData_ComplexGenericField_MultipleTypes() {
+        val text = """
+            data foo {
+                map<string, list<int32>> f1 = 3;
```
victorhurdugaci: Consider adding one more test in which the generic arg is of the type of a `data` type

src/dsl/orbit-codegen/src/main/kotlin/cloud/orbit/dsl/java/TypeIndexer.kt
```diff
@@ -20,14 +16,17 @@ internal class TypeIndexer : AstVisitor() {
         Type("float") to TypeName.FLOAT,
         Type("int32") to TypeName.INT,
         Type("int64") to TypeName.LONG,
-        Type("string") to ClassName.get(String::class.java)
+        Type("void") to TypeName.VOID,
```
victorhurdugaci: Is `void` new? Tests for it? What happens when you use it in a parameter?

cesarbs: It's new, but I hadn't considered the parameter case. I'll remove it and add support for it in a separate change covering all the things.

src/dsl/orbit-codegen-dsl/src/test/kotlin/cloud/orbit/dsl/OrbitFileParserTest.kt
cesarbs: I cover that in `JavaCodeGeneratorTest`. It doesn't make a difference here.

src/dsl/orbit-codegen-dsl/src/main/antlr/cloud/orbit/dsl/Orbit.g4
cesarbs: For 2, I'm keeping `type` since in code we already make the distinction between "type" and "type name" (the latter being the Java Poet concept).

PR #386, opened 2019-05-14
Remove dumphex from logging

olgatsib:
When serialization issues occur the logs get spammed with a lot of garbage data. A single log line stating there was a serialization issue is enough.

### DIFF
--- a/actors/runtime/src/main/java/cloud/orbit/actors/runtime/SerializationHandler.java
+++ b/actors/runtime/src/main/java/cloud/orbit/actors/runtime/SerializationHandler.java
@@ -144,7 +144,6 @@ public class SerializationHandler extends HandlerAdapter
         catch (Throwable e)
         {
             logger.error("Error deserializing message", e);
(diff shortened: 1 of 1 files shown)


actors/runtime/src/main/java/cloud/orbit/actors/runtime/SerializationHandler.java
```diff
@@ -144,7 +144,7 @@ public void onRead(HandlerContext ctx, final Object msg)
         catch (Throwable e)
         {
             logger.error("Error deserializing message", e);
-            logger.error(InternalUtils.hexDump(32, message.getRight(), 0, message.getRight().length));
+            logger.debug(InternalUtils.hexDump(32, message.getRight(), 0, message.getRight().length));
```
johnou: Better to wrap this line in `isDebugEnabled` or the JVM will anyway create a hex dump from the payload and generate unnecessary garbage.

olgatsib: Done

PR #393, opened 2019-05-28
Disable class info for resource only scan.

johnou:
Our new auto-balancing actor system caused Orbit to invoke cloud.orbit.actors.runtime.DefaultClassDictionary#ensureLoaded which causes services configured with a small heap to OOM.

### DIFF
--- a/actors/core/src/main/java/cloud/orbit/actors/runtime/DefaultClassDictionary.java
+++ b/actors/core/src/main/java/cloud/orbit/actors/runtime/DefaultClassDictionary.java
@@ -97,7 +97,7 @@ public class DefaultClassDictionary
             }
 
             long start = System.currentTimeMillis();
-            try (ScanResult scanResult = new ClassGraph().enableClassInfo().scan())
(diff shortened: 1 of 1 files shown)


lukehutch:
See discussion I just posted to 

https://gitter.im/classgraph/Lobby

However in summary, this should probably be:

```java
            try (ScanResult scanResult = new ClassGraph()
                    .whitelistPathsNonRecursive(META_INF_SERVICES_ORBIT_CLASSES)
                    .scan())
            {
                scanResult.getResourcesWithExtension(EXTENSION)
                        .forEachInputStream((Resource resource, InputStream stream) -> {
                            loadClassInfo(resource.getPath(), stream);
                        });
            }
```

or use `.whitelistPaths()` rather than `.whitelistPathsNonRecursive()`, if the resource can be in a subdirectory of `META_INF_SERVICES_ORBIT_CLASSES`.

johnou:
thanks @lukehutch I updated to use whitelistPaths as there are sub folders.

PR #422, opened 2020-04-13
Client-side Zero Down Time addressable draining

brettmorien:


src/orbit-server/src/main/kotlin/orbit/server/mesh/ClusterManager.kt
```diff
@@ -36,8 +35,7 @@ class ClusterManager(
     private val clusterNodes = ConcurrentHashMap<NodeId, NodeInfo>()
     private val nodeGraph = AtomicReference<Graph<NodeId, DefaultEdge>>()
 
-    fun getAllNodes() = clusterNodes.filter { clock.inFuture(it.value.lease.expiresAt) }.values
-
+    fun getAllNodes() = clusterNodes.filter { clock.inFuture(it.value.lease.expiresAt) && it.value.nodeStatus == NodeStatus.ACTIVE }.values
```
JoeHegarty, line 38: Is this always desirable? Are there never cases where we might want to get nodes that are draining?

PR #430, opened 2020-06-08
Client Shutdown Addressable Deactivation

brettmorien:


brettmorien:
Now with 4 modes of deactivation: instant, concurrent, time span and rate limited.

PR #436, opened 2020-07-23
Server ZDT and client improvements

brettmorien:


src/orbit-proto/src/main/proto/orbit/shared/messages.proto
```diff
@@ -10,6 +10,7 @@ message MessageProto {
     NodeIdProto source = 2;
     MessageTargetProto target = 3;
     MessageContentProto content = 4;
+    int64 Attempts = 5;
```
JoeHegarty, line 13: Should this be lower case? Seems like it doesn't match the other fields

src/orbit-client/src/main/kotlin/orbit/client/OrbitClient.kt
```diff
@@ -126,7 +128,9 @@ class OrbitClient(val config: OrbitClientConfig = OrbitClientConfig()) {
             }
 
             // Get first lease
-            nodeLeaser.joinCluster()
+            retry(retryDelay = Duration.ofSeconds(1)) {
```
JoeHegarty, line 131: Should this retry forever or do we need to give up eventually and crash or something?

brettmorien: Fixed it up. Not positive RemoteException is the best way to present connection failure...

PR #439, opened 2020-08-19
Allow actors to throw during deactivation

brettmorien:


src/orbit-client/src/main/kotlin/orbit/client/execution/ExecutionHandle.kt
```diff
@@ -132,14 +132,18 @@ internal class ExecutionHandle(
...
+                        DeferredWrappers.wrapSuspend(
+                            it.method,
+                            instance,
+                            reasonArgs
+                        )
+                    } else {
+                        DeferredWrappers.wrapCall(it.method.invoke(instance, *reasonArgs)).await()
+                    }
+                } catch (ite: InvocationTargetException) {
+                    logger.info { "Exception caught on actor deactivation" }
```
JoeHegarty, line 146: This should be a warning I think. And should write out the exception contents.

brettmorien: I can do that...

PR #445, opened 2021-06-15
fix main method in Hello World example

kenpusney:
- proxy creation using a different API.
- variable name declared as `orbitClient` instead of `orbit`.
- imports includes unnecessary classes but not required ones.


brettmorien:
Thanks for the find and the fix!

